PR #13, opened 2021-07-29
Chore/PrivateRouteHome-C2D-16

Alperdec:
Problem
=======
C2D-16 on Jira. 

Solution
========
Added a private route to /home

Change Summary:
---------------
PrivateRoute home, Route to '/' with new component 'Landingview'
and added some sass changes because the hover jitter was annoying
Dev-Ops
=======
- [ ] webpack-config.js

-----------------------


sample.env
```diff
@@ -1,2 +0,0 @@
-REACT_APP_AUTH0_DOMAIN=
-REACT_APP_AUTH0_CLIENTID=
```
w-gao: Why is this removed?

w-gao (changes requested):
We should probably keep `sample.env` as a template right? Otherwise, this looks good to me.

w-gao (approved):
LGTM!

krishnakpandian (approved):
Everything Looks good and can be merged!

w-gao:
@krishnakpandian Just to confirm, we are doing squash and merge right?

krishnakpandian:
> @krishnakpandian Just to confirm, we are doing squash and merge right?

Yeah it's better if we squash and merge, I just switched it over

PR #21, opened 2021-08-01
Adding logic for resend verification email, and non-verified users route

krishnakpandian:
Problem
=======
Currently, users who have not Verified their email can enter private Routes

Solution
========
* Added a View and checked to see if the user's email is currently verified before displaying page content

Change Summary:
---------------
* Modified private-route.tsx to display the verify view when a user has not verified their email
* Create Verify View
* Created framework for requesting a resend of the verification email
* Created API mock folder for testing

Dev-Ops
=======
- [ ] webpack-config.js

-----------------------


src/views/Verify/index.view.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+const Verify: React.FC = () => {
+  const { user } = useAuth0()
+  return (
+    <>
+      <div className='verify-container'>
+        You are not a verified user
+        <button
+          type='button'
+          onClick={() => resendVerificationEmail(user?.sub)}
```
w-gao, line 14: Is this supposed to POST to the current url? I'm getting a 404 error and it doesn't resend the verification email.

src/views/Verify/index.test.js
```diff
@@ -0,0 +1,14 @@
...
+
+jest.mock("../../utils/Api", () => jest.fn())
+
+Enzyme.configure({ adapter: new Adapter() })
+
+describe('Testing Verify View', () => {
+  it('Verify Page renders without crashing', () => {
+    shallow(<Verify />);
+  });
+});
```
w-gao: EOF

src/auth/PrivateRoute.tsx
```diff
@@ -21,18 +22,24 @@ const PrivateRoute: React.FC<PrivateRouteProps> = ({
...
-    })}
-    path={path}
-    exact={exact}
-    sensitive={sensitive}
-    strict={strict}
-  />
-)
+}: PrivateRouteProps) => {
+  const { user } = useAuth0()
+  const route: React.ComponentType<any> = user?.email_verified
```
w-gao, line 27: I had to logout then login after I verified my email to get through this. I think it's because I made auth0 to [store cache to localstorage](https://github.com/CruzHacks/cruzhacks-2022-website/blob/ed892ec7518baa77282a121b827a29359d8fdd64/src/auth/Auth0ProviderWithHistory.tsx#L34). Perhaps we don't need it to store cache, or can we refresh them somehow?

w-gao (changes requested):
This should mostly work, but I have a few comments. Thanks for doing this!

src/views/Verify/index.view.tsx
krishnakpandian: So this is something we need to do in the backend, I just created a mock function for now to eventually be routed to our backend function. This is something that will be done in our next sprint.

src/auth/PrivateRoute.tsx
krishnakpandian: With auth0 it's caching in localstorage since that's how we have it configured. We can make an endpoint that gets the user's verified status when localstorage is false to double verify. Another option is to just ask people to login and logout after their first login (once verified).  Another option is just to use session storage or store it in state variables and change the auth context. The downside of this approach is after page is on refresh this won't save the current page state and the user would have to relogin (This is how it was implemented for CruzHacks 2020).

w-gao: Looks like you went with the last approach?

w-gao (approved):
LGTM.

PR #22, opened 2021-08-11
Feature/resend verification

krishnakpandian:
Problem
=======
Card 15, 19, 23 [Jira](https://cruzhacks-dev.atlassian.net/jira/software/c/projects/C2D/issues/)

Solution
========
* Needed to add a bit of infrastructure to allow the API to be configured 
* Added Audience
* Configured endpoint function more clearly

Dev-Ops
=======
- [x] Key & Sample value to `.env` & `sample.env` 
- [x] GCP Secret Manager (Both development and production)
- [x] Github Secrets (Added a development and production variable)
- [x] Github Workflows `.github/workflows/dev.yml` & `.github/workflows/prod.yml`
- [x] webpack-config.js

-----------------------
* Still primarily infrastructure oriented
* Github Credentials will need to be added once the API is configured appropriately


src/views/Verify/index.view.tsx
```diff
@@ -1,21 +1,31 @@
...
-import { resendVerificationEmail } from "../../utils/Api"
+import ResendVerification from "../../components/ResendVerification/ResendVerification"
 
 const Verify: React.FC = () => {
-  const { user } = useAuth0()
+  const { user, getAccessTokenSilently } = useAuth0()
+  const [token, setToken] = useState<string>("")
+  const [, setError] = useState<string>("")
+  useEffect(() => {
+    if (user?.email_verified) {
```
w-gao: I think we still need to pass the access token if the user is not verified right? Otherwise they won't be authorized to request a resend.

src/utils/Api.ts
```diff
@@ -1,49 +1,26 @@
...
 /*
 Mock Auth0 Resend Verification Email
 Params: userId -> string | undefined
 Return: n/a
 */
 
-const resendVerificationEmailEndpoint = ""
-export function resendVerificationEmail(userId: string | undefined) {
+const resendVerificationEmailEndpoint =
+  process.env.REACT_APP_RESEND_VERIFICATION_ENDPOINT || ""
```
w-gao: This is fine for now, but when we have more endpoints, we should consider storing the base url as the environment variable and appending the relative endpoints here.

w-gao (commented):
This looks pretty good, but see below for comments. Thanks for adding this!

Alperdec:
Looks good, but I am getting a cors error from localhost:3000 upon testing the resend verification email button. Maybe we should add cors header.

sample.env
```diff
@@ -1,2 +1,4 @@
 REACT_APP_AUTH0_DOMAIN=
 REACT_APP_AUTH0_CLIENTID=
+REACT_APP_AUTH0_AUDIENCE=
+REACT_APP_RESEND_VERIFICATION_ENDPOINT=
```
w-gao:
```suggestion
REACT_APP_AUTHENTICATION_ENDPOINT=
```

src/views/Portal/index.view.tsx
```diff
@@ -18,11 +19,12 @@ const PortalView: React.FC = () => {
           setToken("")
```
w-gao, line 19:
```typescript
    if (user?.email_verified) {
...
```

This needs to be updated as well, I think?

w-gao (approved):

src/views/Portal/index.view.tsx
krishnakpandian: Since this is the portal view, only email verified users should be viewing/using it. Not 100% sure if it's necessary or if there will be any other way to access that view with a non-verified email but we can definitely remove it if you feel it's best to remove it.

w-gao: Ahh, it's fine to keep it this way then.

Alperdec: Yeah we should keep it, I actually think unverified users can access the portal view. I just tried it out, I navigated to "portal" before signing in then I registered a new account, after authenticating I was redirected back to the portal view.

Alperdec (approved):

src/views/Portal/index.view.tsx
Alperdec: unless we're removing the option to navigate to your portal on the main page before logging in

w-gao: @Alperdec is this what you're seeing? If so, this is the Verify page and should be expected. If not, we probably did something wrong. 

image

krishnakpandian: I also just tried replicating your method and was not able to. But the behavior that William commented is what I got and what we should be expecting. Based on web interactions on the final UI design, we can change this code in the future.

Alperdec: Whoops. It's working as expected

PR #23, opened 2021-08-14
Add Button and CoolDownButton components

w-gao:
Problem
=======
First step for [C2D-21](https://cruzhacks-dev.atlassian.net/browse/C2D-21). After clicking the Resend Verification Email button, the button will be locked for 2 mins. 

Solution
========

Change Summary:
---------------
* Add Button and CoolDownButton components

Dev-Ops
=======
- [ ] webpack-config.js

-----------------------


src/components/ResendVerification/ResendVerification.tsx
```diff
@@ -9,12 +10,16 @@ const ResendVerification: React.FC<VerificationEmailProps> = ({
 }: VerificationEmailProps) => (
   <>
     <div className='reset-verification'>
-      <button
-        type='button'
-        onClick={() => resendVerificationEmail(user, token)}
+      <CoolDownButton
+        label='resend'
+        duration={1000 * 120}
```
krishnakpandian: Initially, we said 2 minutes, but after testing I feel like it might be a little long. We can reduce this to maybe 30 seconds so it's a bit nicer UI wise

src/components/Button/CoolDownButton.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+    return () => {}
+  }, [disabled])
+
+  const handleOnClick = () => {
+    setDisabled(true)
+    if (localStorageKey) {
+      localStorage.setItem(localStorageKey, `${Date.now() + duration}`)
+    }
+    onClick()
```
krishnakpandian: Logically this makes a lot of sense. I wonder if an error occurred if we should reenable the button immediately. I'm not 100% what the best design for this scenario is but it's something to think about.

krishnakpandian (commented):
Overall, everything looks good and functions nicely!

krishnakpandian (approved):

src/components/Button/CoolDownButton.tsx
Alperdec: it wouldn't be a bad call, that would be a frustrating situation to deal with. I'm having some trouble foreseeing what kind of errors would occur in this situation though outside of localStorage maybe, what were you thinking?

Alperdec (commented):
Looks good, code is clean and straightforward. Thanks William!

Alperdec (approved):

krishnakpandian:
> Looks good, code is clean and straightforward. Thanks William!

src/components/Button/CoolDownButton.tsx
krishnakpandian: I was thinking maybe if something occurred on Auth0's side but I think it should be fine to leave it as it is. In case there is some exploit that can guarantee an error, this implementation would still avoid people being able to continuously spam. It's probably the most secure as it is now.

w-gao: Ah, good point. Unfortunately we can't do a simple try-catch here, because our `onClick()` functions are most likely async. We might need to pass along a callback function to the `onClick` prop, so that we can re-enable the button in case of an error. 

localStorage could definitely fail as well... so we have to make sure this is still usable if the browser doesn't support / allow localStorage.

krishnakpandian (approved):

Alperdec (approved):
looks good

PR #25, opened 2021-08-19
Feature/resend verification recaptcha

Alperdec:
Problem
=======
What problem or issue does this pull request solve [Jira](https://cruzhacks-dev.atlassian.net/browse/C2D-22)

Solution
========
* 

Change Summary:
---------------
Added recaptcha, added "verified" toggle to unlock resend verification email, added two env variables

Dev-Ops
=======
- [x] Key & Sample value to `.env` & `sample.env` 
- [x] Github Secrets (Added a development and production variable)
- [x] Github Workflows `.github/workflows/dev.yml` & `.github/workflows/prod.yml`
- [x] webpack-config.js

-----------------------


sample.env
```diff
@@ -2,3 +2,5 @@ REACT_APP_AUTH0_DOMAIN=
 REACT_APP_AUTH0_CLIENTID=
 REACT_APP_AUTH0_AUDIENCE=
 REACT_APP_AUTHENTICATION_ENDPOINT=
+REACT_APP_RECAPTCHA_SITE_KEY=
+REACT_APP_RECAPTCHA_VERIFICATION_ENDPOINT=
```
w-gao: EOF

package.json
```diff
@@ -31,6 +32,7 @@
     "jest-junit": "^12.2.0",
     "react": "^17.0.2",
     "react-dom": "^17.0.2",
+    "react-recaptcha": "^2.3.10",
```
w-gao: [`react-recaptcha`](https://www.npmjs.com/package/react-recaptcha/v/2.3.10) doesn't seem to be actively maintained (last publish was 3 years ago), but if we don't run into any limitations we should be good.

w-gao (commented):
This looks like it should work. Will do a thorough testing once I get access to the recaptcha site key. Nice work!

package.json
Alperdec: Hmm, well if we do,  react-google-recaptcha's last update was a year ago and that would be a pretty simple fix. Otherwise I can use google-recaptcha's JS API to make the component

krishnakpandian: That's a good catch, you can use `react-google-recaptcha` as well. Since this is probably part of the site we will require a ReCaptcha it should be okay to leave it as it is. I'll leave the call to you if you do want to update it since this module still supports v2.

krishnakpandian (changes requested):
Everything looks good, make sure to add in the envs to their proper locations. So add this to GCP frontendevs and the recaptchaVerify for the backend, github secrets for the frontend, and webpack-config.js needs to be updated with the new envs as well.

package.json
Alperdec: switched to `react-google-recaptcha`, it's probably not the best idea to use libraries that haven't been updated

public/index.html
```diff
@@ -24,6 +24,7 @@
       Learn how to configure a non-root public URL by running `npm run build`.
     -->
     <title>CruzHacks 2022</title>
+    <script src="https://www.google.com/recaptcha/api.js?onload=onloadCallback&render=explicit" async defer></script>
```
w-gao: Is this still needed for `react-google-recaptcha`?

src/components/ResendVerification/ResendVerification.tsx
```diff
@@ -1,22 +1,55 @@
-import React from "react"
+import React, { useState } from "react"
 import "./ResendVerification.scss"
+import axios, { AxiosRequestConfig, AxiosResponse } from "axios"
+import Recaptcha from "react-google-recaptcha"
```
w-gao: I had to change this to `import ReCAPTCHA from "react-google-recaptcha"` when I first built it, but this seems to work fine now... not sure what was going on there...

w-gao (approved):
This looks great to me! Thanks for switching to `react-google-recaptcha`.

public/index.html
krishnakpandian: We can remove this line since we are using `react-google-recaptcha`

krishnakpandian (changes requested):
Tested it and it seems good still, you just need to update `dev.yml` and `prod.yml`

public/index.html
Alperdec: shoot, no it's not needed

src/components/ResendVerification/ResendVerification.tsx
Alperdec: yeah me neither. I updated it to ReCAPTCHA anyway

krishnakpandian (approved):

PR #28, opened 2021-08-29
Frontend Organizers Route

w-gao:
Problem
=======
Frontend organizers route [Jira C2D-43](https://cruzhacks-dev.atlassian.net/browse/C2D-43).

Solution
========
* Add `admin` prop to PrivateRoute

Change Summary:
---------------
* Added `admin` prop to PrivateRoute so that we can define routes that only people with the organizer role can access.
* Added a test Dashboard view, also renamed `Verify` view to `VerifyView`.

Dev-Ops
=======
- [ ] webpack-config.js

-----------------------

### DIFF
--- a/src/auth/PrivateRoute.tsx
+++ b/src/auth/PrivateRoute.tsx
@@ -1,11 +1,13 @@
 import * as React from "react"
 import { Route } from "react-router-dom"
 import { withAuthenticationRequired, useAuth0 } from "@auth0/auth0-react"
-import Verify from "../views/Verify/index.view"
+import VerifyView from "../views/Verify/index.view"
+import ErrorView from "../views/Error/index.view"
 
 interface PrivateRouteProps {
   component: React.ComponentType<any>
   path: string
+  admin?: boolean
   exact?: boolean
   sensitive?: boolean
   strict?: boolean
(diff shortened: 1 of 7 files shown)


src/auth/PrivateRoute.tsx
```diff
@@ -15,18 +17,34 @@ interface PrivateRouteProps {
...
+
+  let route: React.ComponentType<any> = component
+
+  // redirect if email is not verified
+  if (!user?.email_verified) {
+    route = VerifyView
+  }
+
+  // if admin route, redirect if not admin
+  const namespace = "https://example.com"
```
w-gao: @krishnakpandian I can't seem to find where this came from. This doesn't seem to match the `audience`, and I couldn't find it on Auth0. Is this something that we can access with react-auth0?

krishnakpandian: This is with a hook auth0 supports in terms of using a config. Right now, it seems to default to example.com and not retrieve the audience but I'll try updating to see if I can get it to support using the audience. That's just what Auth0 default hook is, we can probably update the field name to be something easier to parse like `roles` if you think that might be better as well. I'll try having it fixed by tomorrow.

krishnakpandian: Just did some quick testing and looked through some documentation but it seems the namespace is required. Let's just use `https://cruzhacks.com` as our namespace since it doesn't actually refer to the resource it's accessing.

w-gao: Ahh, makes sense. Thanks for looking into this!

krishnakpandian (approved):
Tested and it all seems to be good!

Alperdec (approved):
nice! looks good to me

PR #31, opened 2021-09-06
Chore/mock auth

sidhantas:
Problem
=======
C2D-44

Solution
========
*  Created a manual mock of the useAuth0 method from auth0-react

Change Summary:
---------------
* Created  useAuth0 mock

-----------------------
* If anyone has any requests for other mocks please let me know


__mocks__/@auth0/auth0-react.js
```diff
@@ -0,0 +1,8 @@
+const useAuth0 = () => ({
+  user: "random user",
```
krishnakpandian: Can we update this `user` object to also include fields for
`email_verified`
`email`
`https://cruzhacks.com/roles`
`sub`

krishnakpandian: Along with this, have it so email_verified can either be tested with true or false. https://cruzhacks.com/roles will have an array of values either including the value "Organizer" or "User"

krishnakpandian (changes requested):

__mocks__/@auth0/auth0-react.js
sidhantas: Sorry, I'm kind of confused for the third field, from what I'm reading the object should look like this:
`{
  email_verified: true/false,
  email: "mock email",
  roles: ["Organizer", "User"],
  sub: "mock user"
}`

Correct me if I'm wrong here

krishnakpandian: What you have is correct, but roles will only contain Organizer or User. Auth0 stores roles as an array within the idToken.

krishnakpandian: @sidhantas One other thing to correct is that the title of the field is "https://cruzhacks.com/roles"

sidhantas: Ah, I got it, just had to put the token into jwt.io to figure it out

krishnakpandian (approved):

w-gao (approved):
This looks like it should work. Thanks @sidhantas!

PR #33, opened 2021-09-08
localStorage functions

w-gao:
Problem
=======
Storage functions. [Jira](https://cruzhacks-dev.atlassian.net/browse/C2D-67)

Solution
========
* Create functions to get and store data within localStorage
  * can be associated with a user and/or have a timeout

Change Summary:
---------------
* Add `store` and `retrieve` functions
* Add some tests

Dev-Ops
=======
- [ ] webpack-config.js

-----------------------

### DIFF
--- a/src/utils/Storage.ts
+++ b/src/utils/Storage.ts
@@ -26,8 +26,7 @@ export function store(
 
     window.localStorage.setItem(key, JSON.stringify(data))
   } catch (err) {
-    // eslint-disable-next-line no-console
-    console.error(err)
+    // ignore errors
   }
 
   return value
@@ -69,8 +68,7 @@ export function retrieve(key: string, defaultValue?: any, user?: string): any {
       return data.value
     }
   } catch (err) {
-    // eslint-disable-next-line no-console
-    console.error(err)
+    // ignore errors
   }
 
   return defaultValue
--- a/src/utils/Storage.ts
+++ b/src/utils/Storage.ts
@@ -1,0 +1,77 @@
+/**
+ * Helper functions to store and retrieve cache data.
+ */
+
+/**
+ * Store a value.
+ *
+ * @param key - A unique string.
+ * @param value - The value to store. Can be anything that is JSON serializable.
+ * @param timeout - The number of seconds from now for this value to be valid.
+ *                  If undefined, this value will never expire.
+ * @param user - If set, associate this value with a user.
+ */
+export function store(
+  key: string,
+  value: any,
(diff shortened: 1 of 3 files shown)


src/utils/Storage.ts
```diff
@@ -0,0 +1,77 @@
...
+    const data = {
+      value,
+      user,
+      ...(timeout && { expiration: `${Date.now() + timeout * 1000}` }),
+    }
+
+    window.localStorage.setItem(key, JSON.stringify(data))
+  } catch (err) {
+    // eslint-disable-next-line no-console
+    console.error(err)
```
krishnakpandian: Can we avoid using console if possible, even if it is error. Generally we just want to avoid outputting any console oriented messages to the user directly.

src/utils/Storage.ts
```diff
@@ -0,0 +1,77 @@
...
+    if (data.user && data.user !== user) {
+      return defaultValue
+    }
+
+    if (data.value !== undefined) {
+      return data.value
+    }
+  } catch (err) {
+    // eslint-disable-next-line no-console
+    console.error(err)
```
krishnakpandian: Same comment as above ^ `Comment on lines +29 to +30`

krishnakpandian (commented):
It looks good, I wonder if we can avoid outputting to console if possible.

Alperdec (commented):
looking good, Nice work with documentation. It's very helpful.

krishnakpandian (approved):

src/utils/Storage.ts
w-gao: That makes sense! Will remove them later today.

w-gao: Should we just ignore the errors then?

krishnakpandian: I feel like we should address errors in some way. Do you think it would be better to return it as a promise and reject if it fails in any way? We could also just return null, default, or some error response in this situation for the function calling this (Edit: Realized your code should handle this). If you feel it's best to console.error() go for it, I'm not really sure in this situation but generally we want to limit the amount of info on what our app is experiencing.

w-gao: It makes sense to not log things to console in a production environment. It _might_ be helpful during development for some situations (but probably not in this case).

I think it's fine to ignore them here - we don't really need to return anything at all for the `store` function, and we'll be returning the default for the function below. If these fail for some reason (e.g.: localStorage is not supported), we just won't use localStorage.

PR #38, opened 2021-09-23
Feature/footer component

Alperdec:
Problem
=======
What problem or issue does this pull request solve [Jira](https://cruzhacks-dev.atlassian.net/browse/C2D-71)

Solution
========
*
Created props for icons, and themes. Theme switching is done in-line in Footer.tsx.

Change Summary:
---------------
Large desktop: 
Large desktop

Tablet: 
Tablet

Small Tablet:
small tablet

Phone: 
phone

Small Phone:
small phone
 
Footer-Theme: footer-team

Dev-Ops

======

=
- [ ] webpack-config.js

-----------------------


Alperdec:
Let me know as well if we want hover animations.

src/components/Footer/Footer.scss
```diff
@@ -0,0 +1,53 @@
...
+        display: flex;
+        flex-direction: column;
+        font-family: Source Sans Pro, sans-serif;
+        font-style: italic;
+        @include responsiveness.font($responsive: 2.3vw, $min: 12px, $max: 28px);
+    }
+    &__text--copyright {
+        position: absolute;
+        right: 10%;
+        bottom: 10%;
```
krishnakpandian, line 45: On screens smaller than 350px I would use have the right margin be 0% and have the font-size be 10px to have it cleaner on small screens

src/components/Footer/Footer.test.js
```diff
@@ -0,0 +1,23 @@
...
+    link: "site",
+    Icon: "Icon",
+    label: "label",
+},]
+
+Enzyme.configure({ adapter: new Adapter() })
+describe('Testing Footer', () => {
+  it('Footer renders without crashing', () => {
+    shallow(<Footer socials={socialsMock} theme={themeMock}/>);
+  });
```
krishnakpandian, line 22: Generally, if we are using components with multiple different types of inputs, we want to test a render of different types of states. Most likely it will work but generally this is how conditionally rendered components can be tested without enzyme.

src/components/Footer/Footer.tsx
```diff
@@ -1,8 +1,33 @@
 import * as React from "react"
 import "./Footer.scss"
+import { FooterProps } from "Props/footer/props"
+import { ReactComponent as FooterHill } from "../../images/footerhill.svg"
```
krishnakpandian, line 6: If the `footerProps` are constant between views, it might be better to import it within the component just so the view page is cleaner.

krishnakpandian (changes requested):
This looks really good! Just a couple small comments.

krishnakpandian (commented):
Footer_Line

I'm not sure if this is just dev tools creating this line or if it's something wrong with our code but I found this. When I viewed the page without dev tools it looked fine but this might be something we have to test on development.

src/components/Footer/Footer.scss
Alperdec: can do

Alperdec (commented):
I noticed that as well. I can only recreate it when using the device toolbar.

src/components/Footer/Footer.tsx
Alperdec: Good point

src/components/Footer/Footer.test.js
Alperdec: in this component all that's being rendered conditionally is the number of icons and links and different inline colors. Should I write some tests to make sure the colors are right and the number of icons rendered are correct?

krishnakpandian: I would just test that each theme renders individually and that should be fine since it's mostly UI.

src/images/socialMedia/insta.svg
```diff
@@ -0,0 +1,14 @@
...
+<path d="M98.0217 73.3835C84.6871 73.3835 73.868 83.9289 73.868 96.9277C73.868 109.926 84.6871 120.472 98.0217 120.472C111.357 120.472 122.175 109.926 122.175 96.9277C122.175 83.9289 111.357 73.3835 98.0217 73.3835ZM98.0217 112.2C89.3709 112.2 82.3538 105.361 82.3538 96.9277C82.3538 88.4945 89.3709 81.6552 98.0217 81.6552C106.673 81.6552 113.69 88.4945 113.69 96.9277C113.69 105.361 106.673 112.2 98.0217 112.2Z" fill="current"/>
+<path d="M128.77 72.4535C128.77 75.4889 126.246 77.95 123.131 77.95C120.017 77.95 117.492 75.4889 117.492 72.4535C117.492 69.4175 120.017 66.957 123.131 66.957C126.246 66.957 128.77 69.4175 128.77 72.4535Z" fill="current"/>
+</g>
+<defs>
+<filter id="filter0_d" x="0" y="0.0927734" width="196.007" height="193.652" filterUnits="userSpaceOnUse" color-interpolation-filters="sRGB">
```
w-gao, line 8: Just a note - these ids (e.g.: `filter0_d`, or `paint0_linear` when we have linear gradients) in the SVG are global, so we might need to rename them if it conflicts with others. I don't see a problem here so we don't need to change them, but if the color / filter looks off in the future, that may be why.

w-gao (approved):
I feel like the while line is due to some rounding errors in Chrome - it happens when I zoom in as well. I don't see it on Safari, but I could be wrong. I don't think this is a big deal. If we want to cheat a bit, we could set a blueish background to the home view to hide it, or add a `margin-top: -1px` to the footer component. 

Otherwise this looks great!

src/images/socialMedia/insta.svg
krishnakpandian: That's good to note, we'll definitely track of these throughout the future.

krishnakpandian (approved):

PR #40, opened 2021-09-25
Feature/error

krishnakpandian:
Problem
=======
C2D-32 [Jira](https://cruzhacks-dev.atlassian.net/browse/C2D-32)

Solution
========
* Whenever a user is on a route that doesn't exist, we want to display an error message.

Change Summary:
---------------
* Updated error view to include redirect to home page, and styles to indicate ERROR 404

Dev-Ops
=======
- [ ] webpack-config.js

-----------------------
*No design exists for Error 404 page

### Mobile Mock
Error-mobile-mock

### Tablet Mock
Error-tablet-mock

### Desktop Mock
Error-desktop-mock


src/views/Error/index.scss
```diff
@@ -0,0 +1,26 @@
...
+
+.error-view {
+    height: 100vh;
+    display: flex;
+    flex-direction: column;
+    align-items: center;
+    justify-content: space-evenly;
+    background: linear-gradient(#8296D9,#BFACF0);
+    @include responsiveness.font(10vw, 72px, 144px);
+    .error-view-message {
```
w-gao:
```suggestion
    &__message {
```

Consider using the [BEM](http://getbem.com/naming/) naming convention. e.g.: `&__message` here in SCSS would correspond to the `.error-view__message` class name.

I think this will improve maintainability and avoid naming collisions (not that this will happen in this view, but more likely in the Home view for common names).

src/views/Error/index.scss
```diff
@@ -0,0 +1,26 @@
...
+        color: var(--color-text-primary);
+        font-family: "Lato";
```
w-gao: We should probably have a fallback font. Another suggestion is to put this in [src/stylesheets/_typography.scss](https://github.com/CruzHacks/cruzhacks-2022-website/blob/development/src/stylesheets/_typography.scss), though this is not a necessity.

src/views/Error/index.scss
```diff
@@ -0,0 +1,26 @@
...
+    }
+    .error-button {
+        @include responsiveness.font(4vh, 16px, 64px);
+        transform: scale(1, 1);
+        display: flex;
+        align-items: center;
+        justify-content: center;
+        font-weight: bold;
+    }
+}
```
sidhantas: You probably just missed this but there's no eof here

src/views/Error/index.scss
w-gao: This doesn't seem to be addressed yet, but we can do that later.

w-gao (approved):

src/views/Error/index.scss
krishnakpandian: That's my error, I forgot to make that update. I will do that now!

PR #41, opened 2021-09-25
Email subscription component

w-gao:
Problem
=======
Part of [C2D-27](https://cruzhacks-dev.atlassian.net/browse/C2D-27). Modified from last year's email subscription component. 

Solution
========
* Add EmailSubscription component

Change Summary:
---------------
* Add basic email subscription component from last year

Dev-Ops
=======
- [ ] webpack-config.js

-----------------------


Alperdec (approved):
LGTM

src/components/EmailSubscription/index.tsx
```diff
@@ -0,0 +1,74 @@
...
+const SubmissionStates = {
+  NotSubmitted: 0,
+  Loading: 1,
+  Submitted: 2,
+  Errored: 3,
+}
+
+interface EmailSubscriptionProps {}
+
+export const EmailSubscription: React.FC<EmailSubscriptionProps> = () => {
```
krishnakpandian: Should we keep this empty interface or just leave it as `React.FC`?

krishnakpandian (approved):
Looks Good!

src/components/EmailSubscription/index.tsx
w-gao: Oh yeah, I was planning to define some props, but I don't think we need this anymore.

src/components/EmailSubscription/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+export const EmailSubscription: React.FC = () => {
+  const [email, setEmail] = useState("")
+  const [state, setState] = useState(SubmissionStates.NotSubmitted)
+  const [message, setMessage] = useState("")
+
+  const submitEmail = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault()
+    setState(SubmissionStates.Loading)
+    setMessage("Please wait while your message is being submitted...")
```
krishnakpandian: One thing, I just remembered really late but can we validate that the form data is not an empty string and is valid email [regex](https://www.w3resource.com/javascript/form/email-validation.php).

krishnakpandian (changes requested):

src/components/EmailSubscription/index.tsx
w-gao: Will do!

w-gao: This has been added, but will probably be best to test with #45.

krishnakpandian (approved):

PR #42, opened 2021-09-25
Update responsiveness.screen() mixin

w-gao:
Problem
=======

I find myself having to copy code to apply styling for multiple breakpoints below a certain screen size.

Solution
========
* Update `responsiveness.screen()` mixin to take in one or more breakpoints.

This will copy the block of code for each breakpoint in the final CSS, so it won’t improve efficiency, but it should improve readability a bit.

Change Summary:
---------------

Dev-Ops
=======
- [ ] webpack-config.js

-----------------------


krishnakpandian (approved):
LGTM

PR #45, opened 2021-09-28
Add Landing component to Home view

w-gao:
Problem
=======
Landing component [Jira](https://cruzhacks-dev.atlassian.net/browse/C2D-27).

Solution
========
* Add Landing component to Home view

Change Summary:
---------------
* Add Landing component to Home view
* Modified the Button component to work better with external links
* Also added `modifier` prop to the Button component so we don't have the pass Button-specific class names to use its modifiers (e.g.: `className="landing-component__Button" modifier="primary"` instead of `className="landing-component__Button Button--primary"`)

Dev-Ops
=======
- [ ] webpack-config.js

-----------------------


w-gao:
Styling needs more work. Also have to add the background graphics.

w-gao:
The Email subscription component (#41) needs to be added in, but this should mostly be ready now.

krishnakpandian (commented):
This looks amazing!! Let's wait until we merge the subscription component before merging

src/views/Home/components/Landing/index.tsx
```diff
@@ -1,9 +1,75 @@
...
+              </div>
+              <div className='landing-component__subscriptionContainer'>
+                <EmailSubscription />
+              </div>
+              <div className='landing-component__buttonContainer'>
+                <Button
+                  className='landing-component__button'
+                  modifier='primary'
+                  label='sponsor button'
+                  href='mailto:sponsor@cruzhacks.com'
```
w-gao, line 54: Is this the link we want for this button?

w-gao (commented):
I modified the Button component (see PR description for details) that needs a re-review. I can make that into a separate PR if that needs more discussion.

src/views/Home/components/Landing/index.scss
```diff
@@ -0,0 +1,138 @@
...
+      width: 90%;
+    }
+
+    @include responsiveness.screen(phone, tablet, big-desktop) {
+      width: 55%;
+    }
+  }
+
+  &__text-box {
+    margin-top: 12%;
```
krishnakpandian, line 52: On small-laptop, I think having a margin-top of 0 seems to be a bit more natural but that's only my opinion on it, we can double check with design.

krishnakpandian (approved):
landing-input-box-sizing
This looks amazing overall but I feel like the height of the landing-text-box differs a bit from the figma. Again I think it's fine for the most part but it might be something worth looking into.

krishnakpandian (approved):
LGTM

src/views/Home/components/Landing/index.tsx
krishnakpandian: This is what we have tentatively linked publicly for 2022, so this is good.

PR #46, opened 2021-09-29
Bug/milestone mobile UI

krishnakpandian:
Problem
=======
Chrome emulator, Safari Emulator, Firebase Hosting Emulator + Chrome emulator have incorrectly positioned the UI of the milestones asset.

Solution
========
* Lowering the asset manually to improve usability on mobile.

Change Summary:
---------------
* Updating image to React Component
* Pushing image down 10 pixels to have better accuracy

Dev-Ops
=======
- [ ] webpack-config.js

-----------------------


w-gao (commented):
This still doesn't look right on some displays.

image

A quick fix is to position the SVG absolute to the `header-text`, because its width is changing on different displays based on the width of the text. See [my testing branch](https://github.com/CruzHacks/cruzhacks-2022-website/compare/bug/milestone-mobile-ui-tweaks) for possible changes.

We could also make the width of the `header-circle` div change as the `header-text` div changes, so we can position absolute to `header-circle`.

PR #48, opened 2021-10-03
Feature/team page component

sidhantas:
Problem
=======
Team page

Solution
========
* Created Cards for each organizer
* Organized cards using flexbox

Change Summary:
---------------
* Something else I/we did

Dev-Ops
=======
- [ ] webpack-config.js

-----------------------


src/views/Team/index.view.tsx
```diff
@@ -1,12 +1,23 @@
...
 const TeamView: React.FC = () => (
   <div className='team-view'>
     <NavBar />
-    <p>Team Page Renders</p>
+    <div className='team-view__our-team'>OUR TEAM</div>
+    <div className='team-view__small-message'>
+      Meet our team of creators, thinkers,
+      <br />
+      strategists, and believers, collaborating
+      <br />
```
krishnakpandian: In general, let's avoid using `<br />` tags. While the spacing is mainly is just the line-height, let's keep these a separate divs.

src/components/TeamCard/TeamCard.tsx
```diff
@@ -0,0 +1,57 @@
...
+    <TeamCard name='Eric Duong' title='Marketing' imgsrc={cat} />
+    <TeamCard name='Eliot Winchell' title='Design' imgsrc={cat} />
+    <TeamCard name='Glomarie Ramos' title='Engineering' imgsrc={cat} />
+    <TeamCard name='Khushi Shah' title='Design' imgsrc={cat} />
+    <TeamCard name='Ajay Bhatia' title='Logistics' imgsrc={cat} />
+    <TeamCard name='Brenden Teo' title='Logistics' imgsrc={cat} />
+    <TeamCard name='Shantelle Toh' title='Engineering' imgsrc={cat} />
+    <TeamCard name='Vince Guan' title='Engineering' imgsrc={cat} />
+    <TeamCard name='Jiewen Li' title='Engineering' imgsrc={cat} />
+    <TeamCard name='Michelle Sheu' title='Sponsorship' imgsrc={cat} />
```
krishnakpandian: So this has some repetitive code and once we get images, this file will get more crowded and harder to maintain as a result. I'm going link these files from milestones to give you an idea, how we can simplify this implementation. This is a [props file](https://github.com/CruzHacks/cruzhacks-2022-website/blob/development/src/Props/milestones/milestones.ts) similar to how I would recommend you implement it and the [render file](https://github.com/CruzHacks/cruzhacks-2022-website/blob/development/src/views/Home/components/Milestones/index.tsx). The primary to look at is the .map() to have it scale more easily.

src/components/TeamCard/TeamCard.tsx
```diff
@@ -0,0 +1,57 @@
...
+  name: string
+  imgsrc: string
+  title: string
+}
+
+const TeamCard: React.FC<CardProps> = ({ name, imgsrc, title }) => (
+  <div className='Card'>
+    <img src={imgsrc} alt={name} className='Card__Image' />
+    <p className='Card__Name'>{name}</p>
+    <p className='Card__Title'>{title}</p>
```
krishnakpandian: In general, let's avoid using `<p />` tags here. I believe you are using flex, so things might change when swapping to a div.

src/components/TeamCard/TeamCard.scss
```diff
@@ -0,0 +1,70 @@
...
+  animation: fadeIn 1s;
+  width: 292px;
+  height: 392px;
+  margin-left: 10px;
+  margin-right: 10px;
+  border-radius: 27px;
+  padding-bottom: 10px;
+  background:  
+    linear-gradient(#34486B, #34486B) padding-box,
+    linear-gradient(to right, #BD9BF3, #C0CAFC, #80B0F8) border-box;
```
krishnakpandian, line 15: I'll say let's wait on making this change until after everything is merged in but we should create variables for these.

krishnakpandian (changes requested):
I haven't tested the UI locally, mainly because I am a bit busy tonight but I'll do a full test tomorrow. Overall, for your first UI component, this looks great!!

src/components/TeamCard/TeamCard.tsx
```diff
@@ -0,0 +1,57 @@
...
+    <div className='Card__Logo'>
+      <div className='Card__Logo__LogoBg'>
+        <img src={LiLogo} alt='LinkedIn Logo' />
+      </div>
```
krishnakpandian: We are going to need to have a link to everyone's LinkedIn as a part of this image. Inside the google drive there is a contact list of all the organizers which should contain all the urls. You will probably need to change the interface as well to include that a LinkedIn Url.

src/components/TeamCard/TeamCard.tsx
```diff
@@ -0,0 +1,57 @@
...
+    <TeamCard name='Avani Karkare' title='Marketing' imgsrc={cat} />
+    <TeamCard name='Chinmay Gowdru' title='Marketing' imgsrc={cat} />
+    <TeamCard name='Eric Duong' title='Marketing' imgsrc={cat} />
+    <TeamCard name='Eliot Winchell' title='Design' imgsrc={cat} />
+    <TeamCard name='Glomarie Ramos' title='Engineering' imgsrc={cat} />
+    <TeamCard name='Khushi Shah' title='Design' imgsrc={cat} />
+    <TeamCard name='Ajay Bhatia' title='Logistics' imgsrc={cat} />
+    <TeamCard name='Brenden Teo' title='Logistics' imgsrc={cat} />
+    <TeamCard name='Shantelle Toh' title='Engineering' imgsrc={cat} />
+    <TeamCard name='Vince Guan' title='Engineering' imgsrc={cat} />
```
krishnakpandian: These title's don't seem to be accurate, so we need to update these as well.

src/components/TeamCard/TeamCard.scss
```diff
@@ -0,0 +1,70 @@
+@use "~stylesheets/typography";
+
+.Card {
+  animation: fadeIn 1s;
+  width: 292px;
```
krishnakpandian: I would update this width to be 80-90%, and set a max-width to 292px which will allow this component to scale to the galaxy fold as well.

krishnakpandian (changes requested):

src/Props/TeamCardProps/cards.ts
```diff
@@ -0,0 +1,117 @@
...
+    imgsrc: cat,
+  },
+  {
+    name: "Eliot Winchell",
+    title: "Design",
+    imgsrc: cat,
+  },
+  {
+    name: "Glomarie Ramos",
+    title: "Engineering",
```
krishnakpandian: This is inaccurate

src/Props/TeamCardProps/cards.ts
```diff
@@ -0,0 +1,117 @@
+import { CardProps } from "Props/props"
+import cat from "./cat.jpg"
+
+const cards: CardProps[] = [
+  {
```
krishnakpandian, line 5: Can we start adding in the LinkedIn Urls?

krishnakpandian (approved):
From a UI perspective this is all good, but let's add the [LinkedIn's](https://docs.google.com/spreadsheets/d/1uX-9SCmT18S9Eos4O5H2TcAJ5Hw7a8XxrezhsrsHxH8/edit#gid=0) and fix 1 or 2 items.

PR #50, opened 2021-10-04
Feature/about container

krishnakpandian:
Problem
=======
C2D-26 [Jira](https://cruzhacks-dev.atlassian.net/browse/C2D-26)

Solution
========
* Implemented About Component
* Everything has been merged in together except the header

Change Summary:
---------------
* Updated About.tsx/scss, AboutBox.tsx/scss, other files are from merge conflicts 

Dev-Ops
=======
- [ ] webpack-config.js

-----------------------
* The mobile background image isn't available to export at the moment but the logic exists for implementing it (Edit: Added Asset)

### Desktop
about-desktop

### Mobile 
about-mobile

### Tablet
about-tablet


src/Props/TeamCardProps/cards.ts
```diff
@@ -0,0 +1,139 @@
+import { CardProps } from "Props/props"
```
w-gao, line 1: I think a lot of the changes aren't supposed to be here. You might want to delete these and merge from main.

PR #53, opened 2021-10-04
Fix TeamCard props

w-gao:
Problem
=======

Fix 

This probably has something to do with how typescript is configured in `tsconfig.json` so it didn't break on CI, but it managed to break during development so I'm proposing a fix.

Solution
========
* Fix TeamCard props

Change Summary:
---------------
* Add props validation to TeamCard
* Make `imgSrc` into camelCase to match the rest
* Fix certain linkedIn URLs that are not using HTTPS

Dev-Ops
=======
- [ ] webpack-config.js

-----------------------


krishnakpandian (approved):

krishnakpandian:
Odd, it worked locally for me and passed the GitHub deployment but we can make these changes

PR #60, opened 2021-10-06
Verification View

krishnakpandian:
Problem
=======
C2D-84 [Jira](https://cruzhacks-dev.atlassian.net/browse/C2D-84)

Solution
========
* Cleaned up the styles for the verification page

Change Summary:
---------------
* Updated Verify View, Resend Verification, and Verification Messages

Dev-Ops
=======
- [ ] webpack-config.js

-----------------------
* Responsiveness and no design currently exists for this component

### Mobile
verify-mobile

### Tablet
verify-tablet

### Desktop
verify-desktop


src/views/Verify/VerifyMessages/VerifyMessage.tsx
```diff
@@ -0,0 +1,37 @@
...
+    },
+    {
+      message: "We are excited to see you at CruzHacks 2022!",
+      key: "Line 4",
+    },
+  ]
+  return (
+    <div className='verify-message'>
+      {Messages.map(({ message, key }: MessageProps) => (
+        <div className='verify-message__text' key={key}>
```
w-gao, line 29: Can we use one of our custom fonts for the messages?

w-gao (commented):
This works! Although, I was wondering if we'll get a design for this page later? If not, we could probably use the background for the error page here as well, just to stay consistent with the rest of the app.

src/views/Verify/VerifyMessages/VerifyMessage.tsx
krishnakpandian: Yeah definitely, I'll add some nicer fonts!

krishnakpandian:
I don't think we will get a design for this, since it's a small component and low priority. I'll send a picture using the error background but for now I'm using team page background as our temporary reference

w-gao:
@krishnakpandian Hmm, let's keep using the team page background as is then. Otherwise we'd have to adjust the font color as well. Sorry for the trouble. 

> I just made some edits, before I push, is this more in line in terms of fonts and background-colors? @w-gao verification-new-background


https://github.com/CruzHacks/cruzhacks-2022-website/pull/65#issuecomment-939394635

krishnakpandian:
> @krishnakpandian Hmm, let's keep using the team page background as is then. Otherwise we'd have to adjust the font color as well. Sorry for the trouble.
> 
> > I just made some edits, before I push, is this more in line in terms of fonts and background-colors? @w-gao verification-new-background
> 
> [#65 (comment)](https://github.com/CruzHacks/cruzhacks-2022-website/pull/65#issuecomment-939394635)

No worries at all! It's good for us to try out different styles so feel free point anything like this in the future!

Alperdec (commented):
Looks like it should work. If we go with team-page bg I would like to edit the ReCaptcha box to be dark mode.

@Alperdec I don't believe we can modify the styling of the recaptcha given the modules we are using but that would be nice if possible.

@krishnakpandian let me look into it, I thought I remember it being a prop I can pass

src/views/Verify/index.view.tsx
```diff
@@ -22,7 +24,9 @@ const VerifyView: React.FC = () => {
   return (
     <>
       <div className='verify-view'>
-        <p>You are not a verified user</p>
+        <NavBar theme='home' />
```
Alperdec: If we do the team page theme, can we pass 'team' to navbar?

Alperdec (changes requested):

src/views/Verify/index.view.tsx
krishnakpandian: Yeah let me if that theme works better and if it does I'll change it to that!

krishnakpandian: Tested that and it looks better, this change has been made!

w-gao (approved):

Alperdec (approved):
LGTM!!

PR #62, opened 2021-10-07
Milestone UI - Fix mobile border

w-gao:
Problem
=======
Fix mobile border bug in milestones component [Jira](https://cruzhacks-dev.atlassian.net/browse/C2D-102).

Solution
========
* Refactored milestone component for mobile

Change Summary:
---------------
* Fixed mobile border
* Also addressed some of Navya's comments

Dev-Ops
=======
- [ ] webpack-config.js

-----------------------

* Desktop

* Laptop

* Tablet

* Mobile

* Smallest


src/views/Home/components/Milestones/index.tsx
```diff
@@ -2,37 +2,42 @@ import * as React from "react"
...
-            <div className='milestone__box'>
-              <div className='milestone__stat'> {milestoneStat}</div>
-              <div className='milestone__name'> {milestoneName}</div>
+      <div className='milestones-component__achievements'>
+        {milestoneResults.map(
+          ({ image, milestoneName, milestoneStat, id }: MilestoneProps) => (
+            <div className='milestones-component__milestone' key={id}>
+              <div className='milestones-component__milestone__image'>
+                <img src={image} alt='error' />
+              </div>
```
krishnakpandian, line 26: If we want we can probably make this a `ReactComponent`  in the props file when and that should hopefully help with consistency.

src/Props/props.ts
```diff
@@ -20,11 +20,6 @@ export interface MilestoneProps {
 export interface AboutProps {
   title: string,
   description: string
-} 
-
-export interface NavbarRouteProps {
-  route: string
-  name: string
```
krishnakpandian, line 27: Was this change meant to be included?

src/views/Home/components/Milestones/index.scss
```diff
@@ -1,128 +1,130 @@
...
-  flex-direction: row;
-  justify-content: space-around;
-  flex-wrap: wrap;
+  background: linear-gradient(var(--color-light-pink), var(--color-lavender), var(--color-sky-blue));
+
+  &__container {
+    display: flex;
+    flex-wrap: wrap;
+    justify-content: center;
+    margin: 50px auto;
```
krishnakpandian, line 93: On bigger screens (1400-2000) the spacing between the milestone-header and the milestone-achievements doesn't space around as cleanly so we might want to include a justify-content: space-around. There is another bug available for having this component scale for ~2000 -> 2500 (mainly because only this component is a little awkward on widescreens), so we can save tackle issue for that ticket or we can try fixing it here.

krishnakpandian (changes requested):
Overall it looks really good and clears the main issue! I would maybe look into the desktop version spacing but other than that it looks good. Thank you for your hard-word!

desktop-milestone-scale

src/Props/props.ts
w-gao: Yeah, not sure why there are two `NavbarRouteProps` here so I removed one.

src/views/Home/components/Milestones/index.scss
w-gao: Ah, you had that before but I was trying to match the design for desktop. I just pushed a change that'll use `justify-content: space-around` for bigger screens, so let me know how that looks. 

I can also keep the original version, but I feel like the `milestone-header` and `milestone-achievements` are too far away on desktop (~1440px).

src/views/Home/components/Milestones/index.tsx
w-gao: I tried that but I don't know a way to pass the className to the component. I suppose we could make `image` a type and make the component here with `React.createComponent()`, but that seems to complicate things.

src/views/Home/components/Milestones/index.scss
w-gao: ...and I didn't realize there was another ticket for milestones, but I think this PR pretty much covers it. We can break this into smaller PRs, or keep this one and you can come in and make any additional changes.

src/views/Home/components/Milestones/index.tsx
krishnakpandian: I assume we would be able to pass it in via props but it's not a big deal so I think this is fine.

src/views/Home/components/Milestones/index.scss
krishnakpandian: I think it looks okay, we should probably send a picture of both to design to see which one they like more and confirm from there.

krishnakpandian: We can probably build off this PR for scaling the screen size for larger screens but this is a good refactor!

krishnakpandian (approved):

w-gao:
Merging this in as well, but can adjust the responsiveness for larger screens in subsequent PRs once we get feedback from design.

PR #65, opened 2021-10-10
Having About page scale on larger screen sizes

krishnakpandian:
Problem
=======
C2D-110 [Jira](https://cruzhacks-dev.atlassian.net/browse/C2D-110)

Solution
========
* Set a max-width to prevent stretching on widescreen
* Adjusted font-responsiveness to be smaller on desktop

Change Summary:
---------------
* Adjusted stylesheets AboutBox.scss and index.scss

Dev-Ops
=======
- [ ] webpack-config.js

-----------------------

### DIFF
--- a/src/views/Home/components/About/AboutBox/AboutBox.scss
+++ b/src/views/Home/components/About/AboutBox/AboutBox.scss
@@ -4,7 +4,8 @@
 .about-box {
     z-index: 1;
     border-radius: 33px;
-    width: 80%;
+    width: 70%;
+    max-width: 1500px;
     height: 100%;
     background-color: var(--color-dark-navy-blue);
     margin: 50px 0 50px 0;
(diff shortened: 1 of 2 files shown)


src/views/Home/components/About/AboutBox/AboutBox.scss
```diff
@@ -4,7 +4,8 @@
 .about-box {
     z-index: 1;
     border-radius: 33px;
-    width: 80%;
+    width: 70%;
+    max-width: 1500px;
     height: 100%;
     background-color: var(--color-dark-navy-blue);
     margin: 50px 0 50px 0;
```
w-gao, line 11: Could we also set a min-height to `&__title` below? For example, something like:

```css
        display: flex;
        flex-direction: column;
        justify-content: center;
        min-height: 50px;
```

w-gao (approved):
LGTM! Plus an unrelated suggestion.

src/views/Home/components/About/AboutBox/AboutBox.scss
```diff
@@ -17,11 +18,20 @@
...
-        @include responsiveness.font($responsive: 4.1vw, $min: 35px, $max: 80px);
+        @include responsiveness.font($responsive: 3.8vw, $min: 35px, $max: 80px);
         @include responsiveness.screen(smallest, phone) {
-            @include responsiveness.font($responsive: 7.1vw, $min: 22px, $max: 40px);
+            @include responsiveness.font($responsive: 7.1vw, $min: 20px, $max: 40px);
+            text-align: center;
             padding: 0 26px 0 30px;
         }
+        @include responsiveness.screen(smallest) {
+            padding: 0 0 0 0px;
```
w-gao: Also, this is just `padding: 0`, right?

krishnakpandian: Oh yeah this was probably me just copying from dev tools, I'll update this!

src/views/Home/components/About/AboutBox/AboutBox.scss
krishnakpandian: Yeah I can add that as well, let me test and see how that'll look on smaller screens

krishnakpandian:
I just made some edits, before I push, is this more in line in terms of fonts and background-colors? @w-gao 
verification-new-background

w-gao (approved):
Looks great to me! Thanks for the changes @krishnakpandian.

PR #66, opened 2021-10-10
Milestones responsiveness

krishnakpandian:
Problem
=======
C2D-112 [Jira](https://cruzhacks-dev.atlassian.net/browse/C2D-112)

Solution
========
* Updated the scaling and spacing on larger screens to support widescreen devices

Change Summary:
---------------
* Updated Milestones stylesheets

Dev-Ops
=======
- [ ] webpack-config.js

-----------------------
### Widescreen (2000 - 2500px)
milestones-widescreen

### DIFF
--- a/src/views/Home/components/Milestones/index.scss
+++ b/src/views/Home/components/Milestones/index.scss
@@ -189,7 +189,7 @@
       width: 180px;
     }
 
-    &__milestone-img {
+    &__img {
       margin-right: 10px;
       width: 70%;
       float: right;
--- a/src/views/Home/components/Milestones/index.scss
+++ b/src/views/Home/components/Milestones/index.scss
@@ -52,7 +52,7 @@
     }
 
     &__milestone {
-      &__image {
+      &__image-container {
         display: flex;
         align-items: center;
         justify-content: center;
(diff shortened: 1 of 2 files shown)


src/views/Home/components/Milestones/index.scss
```diff
@@ -178,16 +185,20 @@
...
-    &__image {
+    &__image-container {
       width: 180px;
+    }
 
-      > img {
-        margin-right: 10px;
-        width: 75%;
-        float: right;
+    &__milestone-img {
```
w-gao: I think you meant to do this?

```suggestion
    &-img {
```

Or

```suggestion
    &__img {
```
and change the class name to `milestones-component__milestone__img` below.

src/views/Home/components/Milestones/index.tsx
```diff
@@ -21,8 +21,12 @@ const Milestones: React.FC = () => (
         {milestoneResults.map(
           ({ image, milestoneName, milestoneStat, id }: MilestoneProps) => (
             <div className='milestones-component__milestone' key={id}>
-              <div className='milestones-component__milestone__image'>
-                <img src={image} alt='error' />
+              <div className='milestones-component__milestone__image-container'>
+                <img
+                  src={image}
+                  alt='error'
+                  className='milestones-component__milestone-img'
```
w-gao: See above.

```suggestion
                  className='milestones-component__milestone__img'
```

w-gao (commented):
This definitely looks better now! Although, the class name for the image in the tsx doesn't match the css currently, so that needs to be changed.

src/views/Home/components/Milestones/index.tsx
krishnakpandian: Good catch, I think I placed this in the wrong scope initially. Thanks for catching this!

w-gao (approved):
Awesome, this looks great now!

PR #67, opened 2021-10-10
Fix hills graphics in the Footer component

w-gao:
Problem
=======

The hills in Footer aren't tall enough on mobile [Jira](https://cruzhacks-dev.atlassian.net/browse/C2D-114)

(Image from Alper)

Solution
========
* Exported another SVG with an aspect ratio more suitable for mobile
* Added `preserveAspectRatio="none"` to both SVGs so they will stretch when necessary

Change Summary:
---------------
* Fix footer hills on mobile
* Also remove the while line between Milestones and Footer
* Reindented the SCSS with 2 spaces

Dev-Ops
=======
- [ ] webpack-config.js

-----------------------

### DIFF
--- a/src/components/Footer/Footer.scss
+++ b/src/components/Footer/Footer.scss
@@ -40,7 +40,7 @@
     position: relative;
     display: flex;
     flex-direction: column;
-    font-family: Source Sans Pro, sans-serif;
+    font-family: "Source Sans Pro", sans-serif;
     font-style: italic;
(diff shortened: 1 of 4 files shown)


krishnakpandian (approved):
I think the changes look fine. I'll approve this but can we get confirmation from @Alperdec that this is functioning on his mobile device

Alperdec (approved):
Seems to do the trick, thank you William!

PR #68, opened 2021-10-10
fixed gaps for phone

sidhantas:
Problem
=======
C2D-103

Solution
========
*  Change gap to margins and media queries

Change Summary:
---------------
* Something else I/we did

Dev-Ops
=======
- [ ] webpack-config.js

-----------------------


src/components/TeamCard/TeamCard.scss
```diff
@@ -6,8 +6,10 @@
...
   min-height: 392px;
   max-height: 500px;
-  margin-left: 10px;
-  margin-right: 10px;
+  margin: 3em 2.5em 0;
+  @media (max-width: 320px) {
+    margin: 3em 10px 0;
+  }
   border-radius: 27px;
   padding-bottom: 10px;
```
krishnakpandian, line 15: We really want to avoid mixing padding and margin on the same component because we can get some weird effects

src/components/TeamCard/TeamCard.scss
```diff
@@ -6,8 +6,10 @@
   max-width: 292px;
   min-height: 392px;
   max-height: 500px;
-  margin-left: 10px;
-  margin-right: 10px;
+  margin: 3em 2.5em 0;
+  @media (max-width: 320px) {
+    margin: 3em 10px 0;
+  }
```
krishnakpandian, line 13: I think on mobile we should have `margin:0; margin-bottom:50px` and on smallest the `margin: 3em 10px 0;` (we should be indicating all four margins instead of 3) and that seems to work nicely on mobile devices. Right now it's crunching it to scale similar to the galaxy fold but the screen width is making it extremely awkward. Also let's use the `responsiveness.screen(smallest)` instead of @media just so we are all adhering to the same viewport changes.

krishnakpandian:
I would remove this for mobile view because it creates an awkward gap. https://github.com/CruzHacks/cruzhacks-2022-website/blob/ece4bf4bbcf34ad9e0934943907ca4cccebd6ace/src/views/Team/index.scss#L24

src/Props/teamCardProps/cards.ts
```diff
@@ -27,14 +27,20 @@ const cards: CardProps[] = [
...
+    imgSrc: cat,
     LinkedInLink: "https://www.linkedin.com/in/tracy-a-rocha/",
   },
   {
     name: "Katie Allen",
-    title: "Director of Logistics",
+    title: "Director of Sponsorship",
     imgSrc: cat,
     LinkedInLink: "https://www.linkedin.com/in/kaitlynallen/",
   },
```
krishnakpandian: Almost everyone has posted their profile picture into google drive so let's start adding those as well so we can crop or change those images if needed

krishnakpandian (changes requested):
Just a couple of changes, right now it's primarily scaling everything as the galaxy fold which is making it awkward for mobile devices. The chrome emulator does seem accurate if you want to test it on there

krishnakpandian (approved):

PR #69, opened 2021-10-10
Bug/navbar format

Alperdec:
Problem
=======
What problem or issue does this pull request solve [Jira C2D-111](https://cruzhacks-dev.atlassian.net/browse/C2D-111)
[Jira C2D-106](https://cruzhacks-dev.atlassian.net/browse/C2D-106)

Solution
========

Change Summary:
---------------

Dev-Ops
=======
- [ ] webpack-config.js

-----------------------


krishnakpandian (approved):
LGTM

src/components/NavBar/NavBar.tsx
```diff
@@ -8,7 +8,7 @@ import NavbarRoutes from "Props/navbar/navbar"
 const NavThemes = {
```
krishnakpandian, line 9: I know this wasn't part of the original plan but I've been playing around and I believe might be helpful just a background safety measure. But we can add the logout button if the isAuthenticated only and have no login. This is just in case someone logs in and we need a way to log them out.

krishnakpandian (changes requested):
One other addition that might be helpful after I have doing some testing regarding previous logic. I would reference our old code from our commit history.

sidhantas:
So, I'm not sure if it's too late for a change like this, but I was thinking that instead of making the navbar become a hamburger when the viewport becomes too small, we should make it so that it becomes like that when the width < height, cuz I feel that at that point the navbar buttons become to small to comfortably use, especially if they're a tablet user.

sidhantas (approved):
LGTM otherwise, although I can't test for IPhone

krishnakpandian:
> So, I'm not sure if it's too late for a change like this, but I was thinking that instead of making the navbar become a hamburger when the viewport becomes too small, we should make it so that it becomes like that when the width < height, cuz I feel that at that point the navbar buttons become to small to comfortably use, especially if they're a tablet user.

I think this makes sense to me. I think for this PR we'll let it pass but I'll definitely check in with design about this functionality for maybe a future small bug

Alperdec:
> So, I'm not sure if it's too late for a change like this, but I was thinking that instead of making the navbar become a hamburger when the viewport becomes too small, we should make it so that it becomes like that when the width < height, cuz I feel that at that point the navbar buttons become to small to comfortably use, especially if they're a tablet user.

That's a good point sid. I don't have a tablet so I haven't personally tested it out, but I could see how including tablets in the mobile design would make sense

src/components/NavBar/NavBar.tsx
```diff
@@ -36,6 +38,16 @@ const NavBar: React.FC<NavProps> = ({ theme }: NavProps) => {
...
     return () => window.removeEventListener("resize", handleResize)
   }, [])
+  const logoutButton = (
+    <button
+      className='NavBar__logout'
+      type='button'
+      onClick={() => logout({ returnTo: window.location.origin })}
+      style={{ textShadow: styling.textShadow }}
+    >
+      logout
```
krishnakpandian: Can we capitalize this for consistency!

Alperdec:
This latest commit is a change that follow's sid's suggestion. Check it out let me know what you think. 
There's two smallish issues I have with this.  On a tablet I wonder if centering the menu options is not the call. It may be cumbersome to reach towards the middle of the screen to navigate to a different page. This is an ipad pro:
image.
Second, the design seems to look good for all devices except for the nest hub: 
image
.

Alperdec:
In the nest hub's case, the width is still larger than the height, however it is small enough for the css breakpoint to take over. So it becomes a combination of both

krishnakpandian (approved):
One small change but it looks good!

src/components/NavBar/NavBar.tsx
Alperdec: No problem

krishnakpandian (changes requested):
One other thing in the .map() let's add a key={name}

krishnakpandian (approved):

PR #70, opened 2021-10-11
added waves and changed background

rvaidun:
Problem
=======
Added waves to 404 page
Changed background to match figma

Solution
========
* 

Change Summary:
---------------
* Something else I/we did

Dev-Ops
=======
- [ ] webpack-config.js

-----------------------


src/views/Error/index.view.tsx
```diff
@@ -11,7 +14,15 @@ const ErrorView: React.FC = () => {
   }
   return (
     <div className='error-view'>
-      <div>
+      <div className='error-view__wavesdiv'>
```
krishnakpandian: Can avoid including `div` in  the className

src/views/Error/index.view.tsx
```diff
@@ -11,7 +14,15 @@ const ErrorView: React.FC = () => {
...
+        <img src={Waves} alt='waves' className='error-view__waves' />
+        <img
+          src={BiggerWaves}
+          alt='wavesbig'
+          className='error-view__wavesbig'
+        />
+      </div>
+      <div className='error-view__messagediv'>
```
krishnakpandian: Same as before rename to avoid naming container with `div`

src/views/Error/index.view.tsx
```diff
@@ -1,6 +1,9 @@
 import * as React from "react"
 import "./index.scss"
 import { useHistory } from "react-router-dom"
+import Waves from "images/error/waves.svg"
+import BiggerWaves from "images/error/waves_dropshadow.svg"
```
krishnakpandian: Can we try importing these as ReactComponents instead of images

krishnakpandian (changes requested):
This looks really good! Just a couple of quick changes!

sidhantas:
I'm not sure if this is just me, but the bottom wave on the error page looks kind of pixelated, it might b the shadow but here's a screenshot
image

krishnakpandian:
> I'm not sure if this is just me, but the bottom wave on the error page looks kind of pixelated, it might b the shadow but here's a screenshot image

I definitely am now able to see it on the big wave, it might be from how it is scaled from the svg, I think we should try cleaning that up as well if possible

krishnakpandian (approved):

PR #72, opened 2021-10-12
Localstorage silent auth still causing issues

krishnakpandian:
Problem
=======
Localstorage was supposed to solve silent authentication login but there still seems to be issues with it not necessarily maintaining state.

Solution
========
* Reimplemented memory for better security and better functionality and same bug

Change Summary:
---------------
* Authentication is still failing on a live site unfortunately

Dev-Ops
=======
- [ ] webpack-config.js

-----------------------


w-gao (approved):
This is unfortunate..

PR #74, opened 2021-10-12
Updating Date

krishnakpandian:
Problem
=======
C2D-120 [Jira](https://cruzhacks-dev.atlassian.net/browse/C2D-120)

Solution
========
* Updating landing with date and location

Change Summary:
---------------
* Updating landing index.tsx

Dev-Ops
=======
- [ ] webpack-config.js

-----------------------
* Design recommended these changes
## Desktop
landing-text-desktop
## Tablet
landing-text-tablet
 ## Mobile
landing-text-mobile

### DIFF
--- a/src/views/Home/components/Landing/index.tsx
+++ b/src/views/Home/components/Landing/index.tsx
@@ -38,10 +38,7 @@ const Landing: React.FC = () => {
                 innovation.
               </div>
               <div className='landing-component__dates'>
-                &#47;&#47; Dates: January 14-16, 2022
-              </div>
-              <div className='landing-component__location'>
-                &#47;&#47; Location: Online
+                January 14-16, 2022 &#160;&#183;&#160; Virtual
               </div>
               <div className='landing-component__subscriptionContainer'>
                 <EmailSubscription />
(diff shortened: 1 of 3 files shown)


src/views/Home/components/Landing/index.tsx
```diff
@@ -37,11 +37,8 @@ const Landing: React.FC = () => {
                 problems, pursue inclusion in tech, and kindle the spirit of
                 innovation.
               </div>
-              <div className='landing-component__location'>
-                &#47;&#47; location coming soon
-              </div>
               <div className='landing-component__dates'>
-                &#47;&#47; dates coming soon
+                January 14-16, 2022 &#160;&#183;&#160; Virtual
```
w-gao: Just to confirm, did you intend to put two spaces before and after the bullet point?

Also, I know I wrote the `&#47;` but this is actually getting confusing. I think we could either replace these with the HTML entity names (i.e.: `&nbsp;&middot;&nbsp;`), or just write them as unicode characters like the copyright symbol.

w-gao (approved):
LGTM. Just one small comment that doesn't really need to be changed.

src/views/Home/components/Landing/index.tsx
krishnakpandian: After looking at this code again, I agree with on this and I'll replace the naming! After checking with design, we are intending for it to have 2 spaces before and after.

w-gao (approved):

PR #75, opened 2021-10-12
Removing Portal View Access for first launc

krishnakpandian:
Problem
=======
C2D-118 [Jira](https://cruzhacks-dev.atlassian.net/browse/C2D-118)

Solution
========
* Removing Portal View for Release 1

Change Summary:
---------------
* Adjusting Routes.tsx

Dev-Ops
=======
- [ ] webpack-config.js

-----------------------
* NOTE: This should be merged in only once the production configs have been tested

### DIFF
--- a/src/routes/Routes.tsx
+++ b/src/routes/Routes.tsx
@@ -3,7 +3,7 @@ import { Route, Switch } from "react-router-dom"
 import { useAuth0 } from "@auth0/auth0-react"
 import PrivateRoute from "../auth/PrivateRoute"
 import HomeView from "../views/Home/index.view"
-import PortalView from "../views/Portal/index.view"
+// import PortalView from "../views/Portal/index.view"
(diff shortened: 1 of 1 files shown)


w-gao (approved):
This might be inconvenient when we develop the portal view, but I can't think of a better solution without overcomplicating things.

krishnakpandian:
@w-gao Don't worry we'll have another PR reverting this after deploying to prod for the dev branch. I mainly just want this in the production environment because Auth0 will be active so realistically anyone could access this route and I'd prefer if it would be hidden.

I do agree I wish I knew a better way of handling management of the 2 branches so we don't need to have another PR afterwards but I can look into that as well.

w-gao:
@krishnakpandian OK cool, having another PR afterwards works. I don't think it's worth spending time to do this in another way, since this is like one of the only times we need to do this.

PR #79, opened 2021-10-13
Fixing main page for safari users

krishnakpandian:
Problem: 
After Viewing Safari, I noticed some styles weren't be applied the same as they were on chrome. Navbar positioning is inaccurate and the arrow for mailing was getting sliced for safari (desktop).

Solution
========
* Modified styles with flex that were having issues

Change Summary:
---------------
* Updated navbar and emailing component to fix the position and sizing

Dev-Ops
=======
- [ ] webpack-config.js

-----------------------
* This rehandled a previous bug and I would like to have this validated on multiple screens if possible

### Before on Safari
Screen Shot 2021-10-12 at 10 45 24 PM
### After on Safari
Screen Shot 2021-10-12 at 10 55 53 PM

### DIFF
--- a/src/components/EmailSubscription/index.scss
+++ b/src/components/EmailSubscription/index.scss
@@ -27,7 +27,7 @@
     width: 12%;
     min-width: 40px;
     padding: 8px;
-
+    display: flex;
     background: #FFFFFF;
     border: none;
     border-radius: 10px;
--- a/src/components/NavBar/NavBar.scss
+++ b/src/components/NavBar/NavBar.scss
@@ -6,10 +6,9 @@
   display: flex;
   flex-wrap: wrap;
   align-items: center;
-  justify-content: right;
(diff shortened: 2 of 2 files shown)


rvaidun (approved):

rvaidun (commented):
Changes look good. Not really needed but maybe we could add some kid of themeing to the tab bar on safari.
Something like this:
https://www.amitmerchant.com/get-your-website-ready-tab-bar-theming-of-safari-15/

krishnakpandian:
@rvaidun This is something we can definitely do in the future. We probably have to check in with design for this but it's doable. I believe I previously implemented logic for this in our public folder to the default theme so we can work off that if we do plan on changing that in the future.

w-gao (approved):
Tested and this seems to fix the navbar and the arrow.

PR #85, opened 2021-10-18
Add Portal container logic

w-gao:
Problem
=======
Portal container logic [Jira](https://cruzhacks-dev.atlassian.net/browse/C2D-117)

Solution
========
* Added
  * function to fetch application status from the API (`application/checkApp` endpoint)
  * logic to render different components based on response

Change Summary:
---------------
* Something else I/we did

Dev-Ops
=======
- [ ] webpack-config.js

-----------------------

### DIFF
--- a/src/views/Portal/index.view.tsx
+++ b/src/views/Portal/index.view.tsx
@@ -44,6 +44,7 @@ const PortalView: React.FC = () => {
               break
             case "pending":
               setStatus(AppStatus.Pending)
+              store("applicationStatus", AppStatus.Pending, 12 * 60 * 60) // store for 12hrs
               break
             default:
               setStatus(AppStatus.Errored)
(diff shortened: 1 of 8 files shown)


src/views/Portal/index.view.tsx
```diff
@@ -1,8 +1,65 @@
...
+              break
+            case "accepted":
+              setStatus(AppStatus.Accepted)
+              store("applicationStatus", AppStatus.Accepted, 12 * 60 * 60) //  store for 12hrs
+              break
+            case "rejected":
+              setStatus(AppStatus.Rejected)
+              store("applicationStatus", AppStatus.Rejected, 12 * 60 * 60) // store for 12hrs
+              break
+            case "pending":
```
krishnakpandian, line 45: It might be okay to store this as well since once they submit it won't change until after release

krishnakpandian (approved):
LGTM!

src/views/Portal/index.view.tsx
w-gao: that makes sense

w-gao:
Merging this in but we can make additional changes as the design updates.

PR #86, opened 2021-10-21
New footer for Home view

w-gao:
Problem
=======
Implement the new footer [Jira](https://cruzhacks-dev.atlassian.net/browse/C2D-100).

Solution
========
* Add graphics for the new footer for the Home view

Change Summary:
---------------

Dev-Ops
=======
- [ ] webpack-config.js

-----------------------


krishnakpandian (commented):
It looks really good!! I did notice a couple things via the emulator.

This is from the emulator and from my mobile testing I didn't see any of these issue but I do have a screen larger than the one where this is occurring on but I am seeing that it is causing a scrollbar into white space. I'm not sure if this is just an emulator issue because I don't think that the container is exceeding into the whitespace but this hasn't existed on previous builds either.

footer-scrolling

Also random since I'm not actually seeing this on my phone but I think this may have created some chrome calculation issue but it's not actually occurring for milestones? Edit: Actually after relooking, I am seeing this on the prod environment chrome emulator but it is displaying accurately on mobile.

milestone-spacingincorrect

src/components/Footer/Footer.scss
```diff
@@ -2,12 +2,36 @@
...
+      padding-top: 70px;
+
+      @include responsiveness.screen(smallest) {
+        padding-top: 15px;
+      }
+
+      @include responsiveness.screen(phone, tablet) {
+        padding-top: 40px;
+      }
+    }
```
krishnakpandian: I would create a separate for tablet, for phone I would have the padding be a little bit smaller ~(25-30px) for phone

w-gao:
@krishnakpandian Hmm, I wasn't able to reproduce the behavior in the first screenshot, with the large white space at the bottom. Do you happen to remember how it looks like that? I'll also do more testing later this evening. 

I did notice the milestone icon bug on the emulator and my phone though.

krishnakpandian:
@w-gao For the first bug, I was just using the emulator on the iPhone 5/SE. The second bug maybe you can look at this bug again next sprint (I'm not really sure why this is occurring inaccurately since it is correct on my phone for both chrome and safari) but we can take a look at this again

w-gao:
@krishnakpandian This is what I see using the emulator on iPhone 5/SE, which looks fine? :




& yeah, I can take the milestone bug next sprint. It looks like that on very small screens.

krishnakpandian (approved):

krishnakpandian:
@w-gao I just ran it again and I'm not seeing that anymore. Not sure why I was seeing that to begin with but it looks good.

PR #89, opened 2021-10-28
Milestones: Fix mobile icons

w-gao:
Problem
=======
Mobile UI for milestones component is running into issues [Jira](https://cruzhacks-dev.atlassian.net/browse/C2D-127)

Solution
========
* Don't limit the height for imgContainer on mobile, because the height is likely going to be larger than 80% of screen

Change Summary:
---------------

Dev-Ops
=======
- [ ] webpack-config.js

-----------------------

- Before:

- After:

On real phone

- Before:

- After:

This shouldn't affect anything else but please check other screen sizes too!

### DIFF
--- a/src/views/Home/components/Milestones/index.scss
+++ b/src/views/Home/components/Milestones/index.scss
@@ -52,6 +52,8 @@
     }
 
     &__milestone {
+      gap: 20px;
+
       &__image-container {
         display: flex;
         align-items: center;
@@ -61,7 +63,7 @@
       }
 
       &__img {
-        width: 100%;
+        width: 95%;
       }
 
       &__box {
--- a/src/views/Home/components/Milestones/index.scss
+++ b/src/views/Home/components/Milestones/index.scss
@@ -56,8 +56,12 @@
         display: flex;
         align-items: center;
         justify-content: center;
-        min-width: 200px;
+        width: 126px;
         height: unset;
+      }
+
+      &__img {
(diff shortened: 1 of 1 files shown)


krishnakpandian (commented):
I think the changes look good, and seem to fix the issue. I want to double check the mobile with design with real quick on the sizing of the images. Maybe it's just me, but the image sizing does seem a little big from 350-450. Design did mention in an initial review that they would like to keep the images <= size of the boxes on mobile but I'll double check with them!

milestones-image-size

w-gao:
@krishnakpandian Ah, yeah, the image sizing does seem big there. I'd be happy to make a change after you get confirmation from design.

krishnakpandian:
@w-gao They checked it out and they like how it looks. They suggest "making the margins between icons and edge and box and edge". I.e: Move the image + box more towards the left in the view.

w-gao:
> @w-gao They checked it out and they like how it looks. They suggest "making the margins between icons and edge and box and edge". I.e: Move the image + box more towards the left in the view.

Awesome! Would this work?

krishnakpandian:
Let me double check with them!

w-gao:
@krishnakpandian I can also add some space between the icon and box, since I just noticed that I made it closer now.

Edit: (add image)

krishnakpandian (approved):
Looks Good, design approved!

PR #90, opened 2021-10-29
Refine admin route

w-gao:
Problem
=======
Refine admin route [Jira](https://cruzhacks-dev.atlassian.net/browse/C2D-128)

Solution
========
* Implemented button to retrieve analytics

Change Summary:
---------------

I've made the following changes to the dashboard:

* Overall layout
  * I'm using a fixed width in the parent container instead of varying width in the child, so we don't need to worry about width in the children components
  * Also some styling changes
* AddAnnouncement component
  * Use `useRef` instead of storing the input as a state every time the input box changes
    * also clear the text if the announcement is submitted successfully
* AppAnalytics component
  * Implemented button to retrieve analytics
  * (https://github.com/CruzHacks/cruzhacks-2022-backend/pull/49)
* Countdown timer
  * Use `setInterval()` instead of `setTimeout()`, and make sure the timer is cleared when time is up or component unmounts
    * reason: before we're recalculating time as soon as the component updates (i.e.: when we call `setTimeLeft()`), which is a lot of render calls. 

Dev-Ops
=======
- [ ] webpack-config.js

-----------------------


src/views/Dashboard/components/AddAnnouncement/index.view.tsx
```diff
@@ -1,49 +1,66 @@
...
     }
+
     if (message.length <= 0 || message.length >= 200) {
-      setNote("Invalid Message")
+      setNote("Invalid message")
       setDisabled(false)
       return
     }
+
+    // TODO: we need punctuations and line breaks right?
```
w-gao, line 41: This regex seems too strict.. we should at least support some punctuation and line break characters.

krishnakpandian: I think that's fair assessment, we can update it any ascii character x20-x7F and newlines. Let me know if that seems reasonable and I'll make a separate chore for the backend as well.

w-gao: Yeah, that's reasonable.

src/components/AppAnalytics/AppAnalytics.tsx
```diff
@@ -6,50 +6,63 @@ import CoolDownButton from "components/Button/CoolDownButton"
...
-        if (res.status !== 201) {
-          setError("Unable to Retrieve")
+        if (res.status === 200 || res.status === 201) {
+          // the server responded successfully, now check our own status code
+          const { status } = res.data
+          if (status === 404) {
+            setError("There are no applicants :(")
+          } else {
+            setData(res.data.message)
+            setError("")
```
krishnakpandian: Correct me if I'm wrong but is it only responding no applicants if this is a 404 response?

krishnakpandian (approved):
It looks a lot better! Thanks for optimizing the rerenders as well!

src/components/AppAnalytics/AppAnalytics.tsx
w-gao: This should (?) work with https://github.com/CruzHacks/cruzhacks-2022-backend/pull/49, except I have to fix the tests for that PR. I can also close that and update the logic here if you prefer the original status code response.

krishnakpandian: I looked through it and I agree in that both APIs should probably use a similar response convention for docs but let's talk more on that PR on valid response codes.

src/components/Timer/Timer.tsx
```diff
@@ -10,46 +10,49 @@ interface Timer {
...
   const calculateTimeLeft = () => {
-    const difference = +new Date(`1-14-2022`) - +new Date()
-    let timeLeft = {
-      days: "0",
-      hours: "0",
-      minutes: "0",
-      seconds: "0",
-    }
+    // Fri Jan 14 2022 15:00:00 GMT-0800 (Pacific Standard Time)
+    const eventDate = new Date("2022-01-14T15:00-08:00")
```
w-gao, line 14: `new Date("1-14-2022")` doesn't work on iOS, so I'm changing this to a format that _should_ work on everywhere according to: https://stackoverflow.com/a/13363791. 

I tested this on chrome and safari and it seems to work, but not sure about the other browsers.

w-gao: I'm also not sure if CruzHacks starts at 3pm, but it is a simple change if I got it wrong.

w-gao (commented):
@krishnakpandian I made a few more changes after your initial review that may need a re-review. This PR is also a pretty large refactor so please let me know if anything doesn't make sense.

src/components/Timer/Timer.tsx
```diff
@@ -10,46 +10,49 @@ interface Timer {
...
-  })
+    const timer = setInterval(() => {
+      const val = calculateTimeLeft()
+      if (!val) {
+        clearInterval(timer)
+        return
+      }
+
+      setTimeLeft(val)
+    }, 500)
```
w-gao, line 52: I also changed this to half a second to alleviate slight delays when I try to sync the timer between multiple windows.

src/components/Timer/Timer.tsx
krishnakpandian: It should be fine choosing 3PM, 7PM, or 12AM

src/views/Dashboard/components/AddAnnouncement/index.view.tsx
```diff
@@ -1,49 +1,66 @@
...
     if (!/^[a-zA-Z0-9 ]+$/.test(title) || !/^[a-zA-Z0-9 ]+$/.test(message)) {
-      setNote("Alphanumeric Characters Only")
+      setNote("Alphanumeric characters only")
       setDisabled(false)
       return
     }
```
krishnakpandian, line 46: We can update the regex for this now or do this separately later

src/components/AppAnalytics/AppAnalytics.tsx
```diff
@@ -6,50 +6,63 @@ import CoolDownButton from "components/Button/CoolDownButton"
...
           Get Analytics
         </CoolDownButton>
-        {error && <div>Unable to Fetch Data</div>}
+        {error && <div className='app-analytics-component__error'>{error}</div>}
       </div>
       <div className='app-analytics-component__results'>
         <div className='app-analytics-component__result'>
           <div className='app-analytics-component__data'>
-            {data !== "" ? data : "X"}{" "}
+            {data.applicantCount ?? "X"}{" "}
```
krishnakpandian: Seems like this works, but I'm curious since I'm not familiar with this notation on how exactly JavaScript is reading it. I'm assuming it's similar to a ternary

krishnakpandian (approved):
Tested the UI, and it seems good! I've reviewed through most of the changes and they seem good!

src/components/AppAnalytics/AppAnalytics.tsx
w-gao: Yeah, `??` is the [null coalescing](https://indepth.dev/posts/1300/double-question-marks-typescript-3-7-nullish-coalescing) operator, which is pretty cool. It's pretty similar to a ternary, but the difference is below:

```typescript

// displays "X" if count is the empty string
data.applicantCount !== "" ? data.applicantCount : "X"


// displays "X" if count is null or undefined
data.applicantCount ?? "X"

// should be exactly the same as this, since I don't think data.applicantCount would be null
data.applicantCount !== undefined ? data.applicantCount : "X"


// also the || operator
// displays "X" if count is falsy (0, "", null, undefined, etc)
data.applicantCount || "X"
```

In practice, I don't think `data.applicantCount` will ever be `""`. It could be `undefined` (if the key is not found in `data`) or `0` (if there's no applicant for this category). I don't think it'll be `null` either.

`||` and `??` are very similar, except `||` would display `X` if the count is 0 but `??` would just display the number 0. 



I might actually change it to `data.applicantCount !== undefined ? data.applicantCount : "X"` now that I think about it, in case `??` is not supported everywhere.

src/views/Dashboard/components/AddAnnouncement/index.view.tsx
w-gao: Let's update this later, since I don't know the regex that would match ascii characters off the top of my head.

krishnakpandian (approved):

PR #92, opened 2021-11-04
Application portal - background

w-gao:
Problem
=======
Implement background component for the portal [Jira](https://cruzhacks-dev.atlassian.net/browse/C2D-138).

Solution
========
* Added `hills.svg` and `waves.svg` from figma
  * `waves.svg` is set to be stretchable according to the design
  * The background should adjust based on the height of the content
* Also moved `&__box` to ApplicationForm and ApplicationStatus

Change Summary:
---------------
* Add portal background from figma

Dev-Ops
=======
- [ ] webpack-config.js

-----------------------

With #94:


src/views/Portal/components/Background/index.scss
```diff
@@ -1,7 +1,33 @@
...
+        left: 0;
+        bottom: 0;
+        width: 100%;
+        height: calc(100% - 24vw);
+        z-index: -1;
+    }
+
+    &__context {
+        z-index: 1;
+    }
```
krishnakpandian: Should we apply styling here so that the children is starting in roughly the same vertical position? We could also just handle that from child container as well.

krishnakpandian (commented):
It looks great! I do want to wait until we finish styling the form just to see how the container scales vertically but I think it should all be good!

krishnakpandian (approved):

src/views/Portal/components/Background/index.scss
w-gao: Yeah, I can add that! You could also merge this in first if this PR is blocking anyone, since that change will probably happen in the weekend.

krishnakpandian: Sounds good! If we can just clear the merge conflict, we can merge this in and move forward from there!

w-gao: On second thought we should probably handle this in the child? Since what we really want to do is to keep the box in the same position, but the `Section X/X` thing is also a child.

krishnakpandian: That makes sense. There is still one merged conflict, most likely from my branch that I merged but we should be good.

w-gao:
Hmm, I'm assuming the `feature/application-portal` branch is also failing the lint test, since I didn't touch `src/views/Portal/components/FormInputs/Radio/index.scss`. Will resolve this shortly.

krishnakpandian:
Yeah I'm looking at your changes and I don't see the linter error there so it probably existed already on the branch.

krishnakpandian (approved):

PR #98, opened 2021-11-07
Feature/faqcomponent

rvaidun:
Problem
=======
There is no FAQ Component in development

Solution
========
What I did
* Created new FAQ Component
* Created Props for the FAQ Component 

Change Summary:
---------------
* Added new props for FAQ
* FAQ Questions are clickable and on click the answer appears below
* FAQ Component is responsive. should work on all screen sizes

Dev-Ops
=======
- [ ] webpack-config.js

-----------------------


src/Props/props.ts
```diff
@@ -18,7 +18,7 @@ export interface MilestoneProps {
 }
 
 export interface AboutProps {
-  title: string,
+  title: string
   description: string
 }
```
krishnakpandian: Is this related to your changes?

.prettierrc.json
```diff
@@ -6,4 +6,4 @@
     "jsxSingleQuote": true,
     "arrowParens": "avoid",
     "endOfLine": "auto"
-}
```
krishnakpandian, line 9: Is this related to your changes?

src/views/Home/components/FAQ/index.scss
```diff
@@ -0,0 +1,75 @@
+@use "./stylesheets/responsiveness";
+@use "./stylesheets/typography";
+
+.faq-component {
+  margin-top: -1px;
+  background-size: cover;
+  background: linear-gradient(
+    var(--color-light-pink),
+    var(--color-lavender),
+    var(--color-sky-blue)
```
krishnakpandian: Again Make sure these are scaling for all screen sizes. I highly suggest you test out with [chrome devTools](https://developer.chrome.com/docs/devtools/). For example this is how your code will output to a tablet user. 

faqTablet
.

src/views/Home/components/FAQ/index.tsx
```diff
@@ -1,9 +1,85 @@
...
+  const target = e.target as HTMLTextAreaElement
+  const next = target.nextElementSibling as HTMLElement
+  if (next) {
+    if (next.style.display === "block") {
+      next.style.display = "none"
+    } else {
+      next.style.display = "block"
+    }
+  }
+}
```
krishnakpandian, line 20: Having the `useState` hook might be cleaner. Generally it's best to avoid placing SASS inside our React. Main reason, is that SASS overrides html/jsx enforced styles and it harder to abstract whether our frontend is for rendering logic versus styles

src/views/Home/components/FAQ/index.scss
krishnakpandian: Similar issues occur on mobile and smaller devices

src/views/Home/components/FAQ/index.scss
```diff
@@ -0,0 +1,75 @@
...
+    width: 25%;
+    height: 11px;
+    background: rgba(255, 255, 255, 0.73);
+    margin: auto;
+    // @include responsiveness.screen(desktop, big-desktop) {
+    //   width: 20px;
+    // }
+  }
+  &__question {
+    font-family: Lato;
```
krishnakpandian: Add a fallback family in case our custom fonts aren't able to be loaded in a browser.

src/views/Home/components/FAQ/index.scss
```diff
@@ -0,0 +1,75 @@
...
+    font-style: normal;
+    font-weight: 800;
+    font-size: 24px;
+    line-height: auto;
+    color: #ffffff;
+    cursor: pointer;
+  }
+  &__answer {
+    font-family: Lato;
```
krishnakpandian: Add a fallback family in case our custom fonts aren't able to be loaded in a browser.

krishnakpandian: Add a fallback family in case our custom fonts aren't able to be loaded in a browser.

src/views/Home/components/FAQ/index.scss
```diff
@@ -0,0 +1,75 @@
...
+    color: #ffffff;
+    display: none;
+  }
+
+  &__questionsandanswers:after {
+    display: table;
+    clear: both;
+  }
+  &__column {
+    float: left;
```
krishnakpandian: I would look into `display: flex` instead of `float`, `clear`, `table`. I think it will help you make more responsive code and also flex is amazing to use.

krishnakpandian (changes requested):
There's still some things that need to be updated. Make sure this is scaling correctly on all screen sizes. Use chrome dev tools to test this. Lastly, make sure you are following the design very closely. Here is our design versus the figma at 1440px.

* Spacing is very different
* Design Far more Width than the expected design

## Our Output
faqActual

## Expected Output
faqExpected

src/views/Home/components/FAQ/index.scss
```diff
@@ -0,0 +1,75 @@
...
+    width: 90%;
+    @include responsiveness.screen(large-tablet, phone) {
+      width: 50%;
+    }
+  }
+  &__faqtitle {
+    font-family: "Source Sans Pro", sans-serif;
+    font-style: normal;
+    font-weight: bold;
+    @include responsiveness.font($responsive: 4.1vw, $min: 35px, $max: 60px);
```
krishnakpandian, line 30: This needs either a margin-bottom or a line-height attached to this

rvaidun: fixed

src/views/Home/components/FAQ/index.scss
rvaidun: I looked into flex box but I couldn't get it to work as well since the items would always overflow past the box surrounding the component. Also we need to flex with a column which requires a set height for the container. When the items are expanded and closed the desired height changes.

src/views/Home/components/FAQ/index.scss
rvaidun: Added

src/views/Home/components/FAQ/index.scss
rvaidun: Scales a lot better on mobile devices.

.prettierrc.json
rvaidun: Not sure whats changed here. Don't think I removed anything. I thought you need the ending bracket.

src/Props/props.ts
rvaidun: Reverted

src/views/Home/components/FAQ/index.scss
```diff
@@ -0,0 +1,78 @@
+@use "./stylesheets/responsiveness";
+@use "./stylesheets/typography";
+
+.faq-component {
+  margin-top: -1px;
+  background-size: cover;
+  background: linear-gradient(
```
krishnakpandian: I'd recommend pulling in the new changes since there's a new footer. I just want to see how that looks distance wise given the new footer.

src/views/Home/components/FAQ/index.scss
```diff
@@ -0,0 +1,78 @@
...
+    display: table;
+    clear: both;
+  }
+  &__column {
+    vertical-align: middle;
+    float: left;
+    width: 33%;
+    @include responsiveness.screen(smallest, phone, tablet) {
+      width: 100%;
+    }
```
krishnakpandian, line 73: faqClippingLargeTablet

The responsiveness has improved immensely! There still some clippings ~800px width.

src/Props/faq/faq.ts
```diff
@@ -0,0 +1,69 @@
...
+  },
+  {
+    question: "Will CruzHacks be virtual or in-person?",
+    answer:
+      "Due to UCSC, Santa Cruz, and California protocols, CruzHacks will remain virtual for CruzHacks 2022. However, there may be a possibility of having a short in-person event for those in the Santa Cruz county that abide by UCSC protocols and feel comfortable to attend.",
+  },
+  {
+    question: "Does it cost anything to participate",
+    answer: "Cruzhacks is free for all participants!",
+  },
```
krishnakpandian: There is a missing question mark here. We are also missing a question in the middle column

src/Props/faq/faq.ts
```diff
@@ -0,0 +1,69 @@
...
+  {
+    question: "What kind of projects can I make?",
+    answer:
+      "Any project (as technical as you want to get) is acceptable as long as it is part of one of the three main categories: Earth, Health, or Social Justice.",
+  },
+  {
+    question: "Do I have to be a UCSC student",
```
krishnakpandian: Question mark here as well

src/Props/faq/faq.ts
```diff
@@ -0,0 +1,69 @@
...
+    answer:
+      "Yes! We have at least 25 mentors ready to assist anyone on their project varying all subjects and all skill levels.",
+  },
+  {
+    question: "Will there be any workshops? On what topics?",
+    answer:
+      "We have a wide range of workshops, both technical and non-technical, planned including but not limited to: Intro to Web Design, Basics of HTML, Machine learning basics, and Resume/Cover Letter.",
+  },
+  {
+    question: "What prizes can I expect",
```
krishnakpandian: Question mark here as well

src/Props/faq/faq.ts
```diff
@@ -0,0 +1,69 @@
...
+    question: "Will there be people there to help me with my project?",
+    answer:
+      "Yes! We have at least 25 mentors ready to assist anyone on their project varying all subjects and all skill levels.",
+  },
+  {
+    question: "Will there be any workshops? On what topics?",
+    answer:
+      "We have a wide range of workshops, both technical and non-technical, planned including but not limited to: Intro to Web Design, Basics of HTML, Machine learning basics, and Resume/Cover Letter.",
+  },
+  {
```
krishnakpandian, line 63: Could we try fixing the vertical spacing on the third column to match the design a little bit more.

src/views/Home/components/FAQ/index.scss
```diff
@@ -0,0 +1,78 @@
...
+    border: 7px solid rgba(255, 255, 255, 0.75);
+    border-radius: 59px;
+    padding: 10px;
+    display: table;
+    // height: 1400px;
+    width: 80%;
+    @include responsiveness.screen(smallest, phone, desktop, big-desktop) {
+      width: 60%;
+    }
+  }
```
krishnakpandian: We may need to handle the scaling a little more on screen sizes smaller than 350px. Specifically we want to support a screen size larger than the galaxy fold. If we can get support for galaxy fold that would be great.

krishnakpandian (changes requested):
This looks a lot better! Just a couple more small changes.

rvaidun:
Just noticed problems with the new footer. The bottom FAQ questions are not clickable anymore. I suspect it's because the footer SVG overlaps with the FAQ component
Screen Shot 2021-11-08 at 10 20 54 AM

src/views/Home/components/FAQ/index.tsx
```diff
@@ -1,9 +1,85 @@
 import * as React from "react"
 import "./index.scss"
+import FAQInformation from "Props/faq/faq"
+import { FAQProps } from "Props/props"
 
+const col1 = FAQInformation.slice(0, 4)
+const col2 = FAQInformation.slice(4, 8)
+const col3 = FAQInformation.slice(8, 12)
```
krishnakpandian: This should be (4,9) and (9,13). This is why I think we are missing a question in the render.

src/components/Footer/Footer.scss
```diff
@@ -6,7 +6,7 @@
   margin-top: -1px;
 
   &--with-graphics {
-    padding-top: 40vw;
+    padding-top: 60vw;
```
krishnakpandian: I feel like this is creating way too large of a gap (also a bit hard-coded) but it might be better to handle this in a separate bug fix so I would tentatively revert this. I'll check with William on how we can fix this.

src/components/Footer/Footer.scss
```diff
@@ -25,6 +25,7 @@
     }
 
     .footer-component__graphics {
+      float: left;
```
krishnakpandian: Remove this change

src/views/Home/components/FAQ/index.scss
```diff
@@ -0,0 +1,81 @@
...
+    margin: auto;
+    border: 7px solid rgba(255, 255, 255, 0.75);
+    border-radius: 59px;
+    padding: 10px;
+    display: table;
+    // height: 1400px;
+    width: 80%;
+    @include responsiveness.screen(smallest, phone, desktop, big-desktop) {
+      width: 60%;
+    }
```
krishnakpandian: The width + font-size still needs to be refined a little for galaxy fold. So I would update this in general for `responsiveness.screen(smallest)`.

faqSmallest

krishnakpandian (changes requested):
Almost there, just a couple small things!

src/components/Footer/Footer.scss
rvaidun: Sounds good

w-gao: So... the sun gradient is what made the entire thing so tall. I tried to scale it down a little but it wouldn't look good anymore. The original design had it go over the container as well. 





I think we can't prevent it from extending out of the container, but we could add a `pointer-events: none;` to the graphics svg so it wouldn't block our click events. 

We could also separate the sun gradient to its own svg, but that would still extend out of the container.

krishnakpandian (approved):

PR #99, opened 2021-11-07
Application portal - Save form data

w-gao:
Problem
=======
Load/Save application form [Jira](https://cruzhacks-dev.atlassian.net/browse/C2D-137)

Solution
========
* Implement `SaveData()`
* Also switched from localStorage to sessionStorage 

Change Summary:
---------------
* Added interfaces for saved props in props.ts
* Updated applicationForm to support saving on next click
* Updated Debug Button to use sessionStorage
* Updated storage.ts + tests to use sessionStorage

Dev-Ops
=======
- [ ] webpack-config.js

-----------------------


src/components/ApplicationContext/ApplicationContext.tsx
```diff
@@ -133,6 +151,18 @@ export const ApplicationProvider: React.FC = () => {
     }
   }
 
+  const saveData = () => {
+    const data: SavedApplication = {
+      contact: contactFormData,
+      demographic: demographicFormData,
+      shortAnswer: shortAnswerFormData,
+      priorExperience: priorExperienceFormData,
+      connected: connectedFormData,
```
krishnakpandian: This seems to not save the file locally (which is good since this is probably a security risk and we don't want to store that much data in the browser) but do we know have code that explicitly handles this or is it just working?

krishnakpandian (commented):
Tested this and it seems like it works! I did have a question on the functionality of storing data.

src/components/ApplicationContext/ApplicationContext.tsx
```diff
@@ -133,6 +151,18 @@ export const ApplicationProvider: React.FC = () => {
...
+      MLH: mlhFormData,
+    }
+    store("application", data, undefined, user?.email)
```
krishnakpandian: Can we also put a timer on this so it doesn't stay indefinitely in our user's localstorage.

krishnakpandian (changes requested):
Did some more testing to make sure different accounts can't access data and I think we should add a timeLimit to the data persistence. (Maybe like a month)

src/components/ApplicationContext/ApplicationContext.tsx
w-gao: This was definitely an oversight on my part. I did not consider the case with the resume, but it looks like the `File` object serializes to `{}` and doesn't fail when we do `JSON.stringify()` (at least on Chrome), but we should definitely explicitly remove that field.

I'm also noticing that this stores the error messages too, which is OK I guess unless we want to strip those out as well.

src/components/ApplicationContext/ApplicationContext.tsx
w-gao: Yep! A month would be `60 * 60 * 24 * 30` seconds right? Just wanted to make sure...

Also, the timeout is just for our logic, it'll still stay in the localStorage if the user doesn't visit our site after a month. So we should probably clear this after the user submits, and/or after the application deadline.

krishnakpandian: I think that's a good call and for our onSubmit we can handle that

src/components/ApplicationContext/ApplicationContext.tsx
krishnakpandian: We can probably just strip the error messages and file. If you want, we can modify the current interface to extend of the fields that you need for storing and retrieving.

src/components/ApplicationContext/ApplicationContext.tsx
```diff
@@ -133,6 +150,48 @@ export const ApplicationProvider: React.FC = () => {
...
+        major,
+        currentStanding,
+        country,
+      },
+      shortAnswer: { whyCruzHacks, newThisYear, grandestInvention },
+      priorExperience: { firstCruzHacks, hackathonCount, priorExperience },
+      connected: { linkedin, github, cruzCoins, anythingElse },
+      MLH: { conductAgree, tosAgree },
+    }
+    store("application", data, 60 * 60 * 24 * 30, user?.email)
```
krishnakpandian: I think one thing we should do is run the validation functions before saving. For example if a user inputs binary data, that will be stored in localstorage. If you run the validation functions with the onSubmit argument set to False it should validate only fields that must always be true (i.e valid alphanumeric data)

krishnakpandian: I was able to save+load emojis/non standard chars

krishnakpandian (changes requested):
Tested again and everything still seems functional. I think one last change.

krishnakpandian:
As a side note, we may want to store appStatus with a user in localstorage so other users don't steal the appStatus.

src/components/ApplicationContext/ApplicationContext.tsx
w-gao: That makes sense! Will do that after today's meeting.

Should we add a flash message above the form to notify the user if we can't save? It'll probably be useful for our submit function too, if the validation fails for the previous pages. Something like this for example:

(example from the Internet)

krishnakpandian: Something like that could work! Could we also hold of some changes on this for now. I've been thinking about future proofing security for 2023 onwards and I wanted to get your thoughts on switching to session storage. If we do switch to session which we'll talk about at today's meeting, we can probably just run the save function onClick of the next/prev button.

Main reason I was thinking of switching is a scenario like this.

Localstorage: User Saves App-> Never submits app -> Data is never cleared -> user never revisits cruzhacks.com for 1 year -> auth0 is off -> stored personal data can never be cleared (unless user manually clears browsers cache) -> Potential XSS vulnerability in 2023 onward or even in current code base (data is compromised) or this could be a problem to future teams using caching (i.e similar naming conventions)

w-gao: This is definitely a valid concern. I also don't like storing the application in local storage.

Although, if we don't want to use local storage, perhaps we shouldn't store it at all? If we switch to session storage, I assume we would remove the "save" option so we don't confuse our user. But how often would a user navigate away from the "/portal" route (our ApplicationContext) and come back within the session?

I guess another option is to just ask for confirmation before the user navigates away from the portal view, similar to how Google Forms does it, and not bother with saving anything that is not permanent.

We should talk more about this during today's meeting.

src/components/ApplicationContext/ApplicationContext.tsx
```diff
@@ -127,14 +158,58 @@ export const ApplicationProvider: React.FC = () => {
...
+  }
+
   const nextPage = () => {
     if (page < 6) {
+      saveData()
       setPage(page + 1)
     }
   }
```
krishnakpandian: It might be good to maybe save the farthest page visited so we can use that in progress bar on page 0 for a started application. We should probably save this if design thinks it's good and in a separate PR.

krishnakpandian (approved):
LGTM! Thanks for putting in the effort for this story, I know this one had a bit more internal design indecisiveness!

src/components/ApplicationContext/ApplicationContext.tsx
w-gao: Oh yeah, that would be cool. There are still a few other things that I want to address before we merge this (unless we need this by Sunday?).

krishnakpandian: I would like this by Sunday, but if there are things that need to be addressed we should prioritize that.

w-gao:
A few notes:

* This will still save unvalidated inputs (e.g.: emojis) when user goes to the previous page
  * should we just not save in this case?
* Save farthest page visited https://github.com/CruzHacks/cruzhacks-2022-website/pull/99#discussion_r747787149
  * rethink how we save 
* I also wanted to add a prompt when there are unsaved changes, but we can't differentiate between user changes and our changes easily (unless we introduce more state variables)

Snippet that may be useful:
```typescript
  useEffect(() => {
    const confirmLeave = (ev: BeforeUnloadEvent) => {
      ev.preventDefault()
      // eslint-disable-next-line no-param-reassign
      ev.returnValue = "Are you sure?"
    }

    window.addEventListener("beforeunload", confirmLeave)
    return () => {
      window.removeEventListener("beforeunload", confirmLeave)
    }
  }, [])
```

krishnakpandian:
@w-gao 
1. Save unvalidated inputs we can probably run the save function [here](https://github.com/CruzHacks/cruzhacks-2022-website/blob/848168cd18b2aedd23e8ba700e73074ee7cae5be/src/views/Portal/components/ApplicationForm/index.view.tsx#L47) instead on next so it does validation before saving. As far as validating on prev, I'm not sure what's the best approach here. We could just not save (which might be annoying in the case a user does refresh unsaved data) or we could validation telling them they have on previous but that could run into a weird issue where a user wants to go back a page first before updating the current page. I think having saving just as a backup in case a user needs to refresh their tab should be the goal for us since we aren't supporting cross tab support. I'm leaning more towards option 1 in this situation but we let's chat on this some more because I'm not sure what the best choice would be in this situation. 
2. This can probably be handled really quickly in its own chore
3. I like this idea a lot. I think this can also be handled in a story next sprint. 

For now, I think if we just fix saving unvalidated inputs, that should be enough for testing on the staging on branch.

src/Props/portal/application.ts
```diff
@@ -1,10 +1,11 @@
 const AppStatus = {
   Loading: 0,
   NotFound: 1,
-  Pending: 2,
-  Accepted: 3,
-  Rejected: 4,
-  Errored: 5,
+  InProgress: 2,
```
w-gao, line 4: Breaking change

src/components/ApplicationContext/ApplicationContext.tsx
```diff
@@ -127,8 +160,64 @@ export const ApplicationProvider: React.FC = () => {
       .catch(() => setStatus(AppStatus.Errored))
   }, [])
 
+  const savePage = () => {
```
w-gao, line 163: I changed this to save current page only, so we won't save unvalidated inputs in the other pages.

src/views/Portal/components/ApplicationStatus/index.view.tsx
```diff
@@ -28,6 +28,9 @@ const ApplicationStatus: React.FC = () => {
         setStatusMessage("NOT STARTED")
         // setStatusMessage("CLOSED") Uncomment to Close Applications
         break
+      // case AppStatus.InProgress:
```
w-gao, line 31: I'll implement this in a separate PR along with the progress bar update once we get design's approval.

w-gao (commented):
I made some more changes that require a re-review.

w-gao:
> I think having saving just as a backup in case a user needs to refresh their tab should be the goal for us since we aren't supporting cross tab support. I'm leaning more towards option 1 in this situation but we let's chat on this some more because I'm not sure what the best choice would be in this situation.

@krishnakpandian I agree! Option 1 would be sufficient for saving I think, especially when we add a confirmation popup when users have unsaved data.

krishnakpandian:
> > I think having saving just as a backup in case a user needs to refresh their tab should be the goal for us since we aren't supporting cross tab support. I'm leaning more towards option 1 in this situation but we let's chat on this some more because I'm not sure what the best choice would be in this situation.
> 
> @krishnakpandian I agree! Option 1 would be sufficient for saving I think, especially when we add a confirmation popup when users have unsaved data.

Awesome, if we're on the same page, I think this PR has all the requirements we need for now. We can add in the confirmation popup + update farthest page visited for home page next sprint or even this sprint if there's time. Correct me I'm wrong on this!

krishnakpandian (approved):
I tested through it and I think it everything seems good

w-gao:
@krishnakpandian Yep, I think this can be merged. I can take those tickets too, but will probably get it done next sprint.

PR #101, opened 2021-11-08
Creating new checkbox based on design feedback

krishnakpandian:
Problem
=======
Checkbox isn't completely what design had in mind in terms of search functionality. It should still be typing based.

Solution
========
* Create an input box that rescales based on inputs that currently already exists

Change Summary:
---------------
* Created checkbox.tsx, checkbox.scss, and modified components that called checkbox

Dev-Ops
=======
- [ ] webpack-config.js

-----------------------
* UI needs to be refined to be cleaner


Alperdec (approved):
The functionality is great and the checkboxes look a lot better than they did before. The only thing I have to add is maybe we can see if we can style the chosen pronouns differently since it is pretty plain

PR #105, opened 2021-11-08
created onSubmit function, not completely styled

sidhantas:
Problem
=======

Solution
========
*  Created a Submit function for application
* Doesn't have number of pronouns or sexualities

Change Summary:
---------------
* Something else I/we did

Dev-Ops
=======
- [ ] webpack-config.js

-----------------------

### DIFF
--- a/src/views/Portal/components/ApplicationForm/index.view.tsx
+++ b/src/views/Portal/components/ApplicationForm/index.view.tsx
@@ -186,8 +186,10 @@ const ApplicationForm: React.FC = () => {
       </div>
       <div className='application-form-component__box'>
         {renderPage(page)}
-        <div>{successOnSubmit === "submitted" ? "submitted" : ""}</div>
-        <div>
+        <div className='application-form-component__response'>
+          {successOnSubmit === "submitted" ? "submitted" : ""}
+        </div>
+        <div className='application-form-component__response'>
           {successOnSubmit === "error submitting" ? "Could Not Submit" : ""}
         </div>
         <div className='application-form-component__buttons'>
--- a/src/views/Portal/components/ApplicationForm/index.view.tsx
+++ b/src/views/Portal/components/ApplicationForm/index.view.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from "react"
+import React, { useState } from "react"
 import "./index.scss"
 import axios from "axios"
 import { useAuth0 } from "@auth0/auth0-react"
@@ -25,6 +25,7 @@ const ApplicationForm: React.FC = () => {
     setPage,
     prevPage,
     nextPage,
+    accessToken,
     contactFormData,
     setContactFormData,
     demographicFormData,
@@ -37,16 +38,7 @@ const ApplicationForm: React.FC = () => {
     setConnectedFormData,
   } = useApplication()!
 
-  const { getAccessTokenSilently } = useAuth0()
-  const [token, setToken] = useState("no token")
   const [successOnSubmit, setSubmitStatus] = useState("none")
-
-  useEffect(() => {
-    getAccessTokenSilently()
-      .then(ac => setToken(ac))
-      .catch()
-  }, [])
-
   const viewNextPage = () => {
     if (page === ApplicationPages.Contact) {
       if (validateContactForm(contactFormData, setContactFormData, true)) {
@@ -100,7 +92,7 @@ const ApplicationForm: React.FC = () => {
       const bodyData = new FormData()
       bodyData.append("fname", contactFormData.fname)
       bodyData.append("lname", contactFormData.lname)
-      bodyData.append("email", user?.email ? user.email : "none")
+      bodyData.append("email", user?.email ? user.email : "")
       bodyData.append("phone", contactFormData.phone)
       bodyData.append("age", demographicFormData.age)
       bodyData.append(
@@ -148,7 +140,7 @@ const ApplicationForm: React.FC = () => {
         data: bodyData,
         headers: {
           "Content-Type": "multipart/form-data",
-          Authorization: `Bearer ${token}`,
+          Authorization: `Bearer ${accessToken}`,
         },
       })
       if (res.status === 201) {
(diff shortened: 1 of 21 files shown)


src/views/Portal/components/ApplicationForm/index.view.tsx
```diff
@@ -82,7 +94,66 @@ const ApplicationForm: React.FC = () => {
...
+      bodyData.append("phone", contactFormData.phone)
+      bodyData.append("age", demographicFormData.age)
+      bodyData.append("pronounCount", "1")
+      for (let i = 0; i < demographicFormData.pronouns.length; i += 1) {
+        bodyData.append("pronouns[0-N-1]", demographicFormData.pronouns[i])
+      }
+      bodyData.append("sexualityCount", "1")
+      for (let i = 0; i < demographicFormData.sexuality.length; i += 1) {
+        bodyData.append("sexuality[0-N-1]", demographicFormData.sexuality[i])
+      }
```
krishnakpandian: Just as a note: pronounCount should be number items in the array and the data appended is each index in the array. Same for sexuality.

krishnakpandian: It should look something more like this
`bodyData.append("pronounCount", integertostr(demographicFormData.pronouns.length)`
`bodyData.append('pronoun[${i}]', demographicFormData.pronouns[i]`)

src/views/Portal/components/ApplicationForm/index.view.tsx
```diff
@@ -82,7 +94,66 @@ const ApplicationForm: React.FC = () => {
...
+        data: bodyData,
+        headers: {
+          "Content-Type": "multipart/form-data",
+          Authorization: `Bearer ${token}`,
+        },
+      })
+      console.log(res)
+      if (res.status === 201) {
+        setSubmitStatus("submitted")
+      }
```
krishnakpandian: If it's submitted can we
1. Remove saved user profile from localstorage
2. Set status of user to submitted in localstorage and in context
3. Set Page status to page 0

src/views/Portal/components/ApplicationForm/index.view.tsx
```diff
@@ -82,7 +94,66 @@ const ApplicationForm: React.FC = () => {
 
   const saveData = () => {}
 
-  const submitData = () => {}
+  const submitData = async () => {
+    try {
+      const bodyData = new FormData()
+      bodyData.append("fname", contactFormData.fname)
+      bodyData.append("lname", contactFormData.lname)
+      bodyData.append("email", contactFormData.email || "none")
```
krishnakpandian: It might be better to use the auth0 Hook right here to grab the email in case of some issue with the contactFormData

krishnakpandian (changes requested):
Just a couple small changes

src/views/Portal/components/ApplicationForm/index.view.tsx
sidhantas: So, I was able to do number 3, but I think we should save number 1 and 2 for a different pull request because we haven't merged the onSave function yet

src/views/Portal/components/ApplicationForm/index.view.tsx
```diff
@@ -34,6 +37,16 @@ const ApplicationForm: React.FC = () => {
...
 
+  const { getAccessTokenSilently } = useAuth0()
+  const [token, setToken] = useState("no token")
+  const [successOnSubmit, setSubmitStatus] = useState("none")
+
+  useEffect(() => {
+    getAccessTokenSilently()
+      .then(ac => setToken(ac))
+      .catch()
+  }, [])
```
krishnakpandian: You can grab the access token from useApplication() so we don't have to rerequest.

src/views/Portal/components/ApplicationForm/index.view.tsx
```diff
@@ -81,8 +94,71 @@ const ApplicationForm: React.FC = () => {
...
   const saveData = () => {}
-
-  const submitData = () => {}
+  const { user } = useAuth0()
+  const submitData = async () => {
+    try {
+      const bodyData = new FormData()
+      bodyData.append("fname", contactFormData.fname)
+      bodyData.append("lname", contactFormData.lname)
+      bodyData.append("email", user?.email ? user.email : "none")
```
krishnakpandian: Instead of "none" leave it as "" so it fails server side.

krishnakpandian (changes requested):
Very quick change but we should be good!

src/views/Portal/components/ApplicationForm/index.view.tsx
```diff
@@ -81,8 +86,71 @@ const ApplicationForm: React.FC = () => {
...
+        "priorExperience",
+        priorExperienceFormData.priorExperience
+      )
+      bodyData.append("linkedin", connectedFormData.linkedin)
+      bodyData.append("github", connectedFormData.github)
+      bodyData.append("cruzCoins", connectedFormData.cruzCoins)
+      bodyData.append("anythingElse", connectedFormData.anythingElse)
+      const res = await axios({
+        method: "post",
+        url: "http://127.0.0.1:5001/cruzhacks-2022-development/us-central1/application/submit",
```
krishnakpandian: Update URL using ENV

src/views/Portal/components/ApplicationForm/index.view.tsx
```diff
@@ -81,8 +86,71 @@ const ApplicationForm: React.FC = () => {
...
+      bodyData.append(
+        "grandestInvention",
+        shortAnswerFormData.grandestInvention
+      )
+      bodyData.append("firstCruzHack", priorExperienceFormData.firstCruzHacks)
+      bodyData.append("hackathonCount", priorExperienceFormData.hackathonCount)
+      bodyData.append(
+        "priorExperience",
+        priorExperienceFormData.priorExperience
+      )
```
krishnakpandian, line 132: Last thing to add `bodyData.append("file", connectedFormData.resume)`

krishnakpandian (changes requested):

PR #107, opened 2021-11-09
Prevent Footer from blocking click events

w-gao:
Problem
=======
Footer padding is extending out of the container and causing it to block the click events of the FAQ component [Jira](https://cruzhacks-dev.atlassian.net/browse/C2D-149)

See: https://github.com/CruzHacks/cruzhacks-2022-website/pull/98#issuecomment-963385707.

Solution
========
* Prevent footer graphics from blocking click events
  * It'll still extend out of the container but hopefully it wouldn't block the click events anymore

Change Summary:
---------------

Dev-Ops
=======
- [ ] webpack-config.js

-----------------------
* **Still need to test on mobile**

### DIFF
--- a/src/components/Footer/Footer.scss
+++ b/src/components/Footer/Footer.scss
@@ -27,6 +27,7 @@
     .footer-component__graphics {
       position: absolute;
       bottom: 0;
+      pointer-events: none;
       z-index: 0;
     }
   }


krishnakpandian (approved):
Tested and this seems to fix the issue on `feature/faqcomponent`

PR #108, opened 2021-11-09
Feature/application portal new searchbox and checkbox

Alperdec:
Problem
=======
Checkbox and Search box dropdown styling was formerly un-styled

Solution
========
updated styles
made the styling consistent between both

Change Summary:
---------------
* Something else I/we did

Dev-Ops
=======
- [ ] webpack-config.js

-----------------------


krishnakpandian (changes requested):
There were a couple bugs with the rendering order but I decided to clean them up real quick. It looks good! It does show that your checkbox changes are in this file and so I would remove those so you don't have conflicts with your other branch.

krishnakpandian (approved):
Actually I can just close my PR but update the description to include both searchbox and checkbox changes.

PR #116, opened 2021-11-13
fixed math div depreactation notice and Typescript error for API keys

sidhantas:
Problem
=======
[C2D-155](https://cruzhacks-dev.atlassian.net/browse/C2D-155?atlOrigin=eyJpIjoiM2E1N2VlYjQyMGVmNDRiZDg5NzYxMjU1MzQwOWE0N2MiLCJwIjoiaiJ9)

Solution
========
* Fixed API deprecation notice

Change Summary:
---------------
* Updated Responsiveness Font function to use math.div() instead of /
* Updated API functions that use API key to set to empty string if env is NULL

Dev-Ops
=======
- [ ] webpack-config.js

-----------------------

### DIFF
--- a/src/stylesheets/_responsiveness.scss
+++ b/src/stylesheets/_responsiveness.scss
@@ -7,6 +7,8 @@ Sample usage::
    @include responsiveness.screen(phone, tablet, large-tablet) {
    }
  */
+@use 'sass:math';
+
 @mixin screen($breakpoints...) {
   @each $breakpoint in $breakpoints {
     @if ($breakpoint == smallest) {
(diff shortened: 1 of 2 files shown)


src/stylesheets/_responsiveness.scss
```diff
@@ -71,16 +73,16 @@ Sample usage::
...
-  $min-breakpoint: $min / $responsive-unitless * 100;
+  $min-breakpoint: math.div($min, $responsive-unitless * 100);
 
   @media (max-#{$dimension}: #{$min-breakpoint}) {
     font-size: $min;
   }
 
   @if $max {
-    $max-breakpoint: $max / $responsive-unitless * 100;
+    $max-breakpoint: math.div($max, $responsive-unitless * 100);
```
krishnakpandian, line 86: If you test the app, you'll notice that responsiveness is breaking. I think the `* 100` should be in the numerator.

krishnakpandian (changes requested):
Let's test the changes across the app, to make sure it doesn't break our current styles.

src/stylesheets/_responsiveness.scss
```diff
@@ -71,16 +73,16 @@ Sample usage::
 ///  @include font(5vw, 35px, 150px, 50px);
 ///
 @mixin font($responsive, $min, $max: false, $fallback: false) {
-  $responsive-unitless: $responsive / ($responsive - $responsive + 1);
+  $responsive-unitless: math.div($responsive, ($responsive - $responsive + 1));
   $dimension: if(unit($responsive) == "vh", "height", "width");
-  $min-breakpoint: $min / $responsive-unitless * 100;
```
krishnakpandian, line 76: Same as previously stated

krishnakpandian (changes requested):

krishnakpandian (approved):

PR #117, opened 2021-11-14
Updating validation messages for clarity

krishnakpandian:
Problem
=======
Validation Messages don't tell the error descriptively to the user in terms of length

Solution
========
* Fixed validation message text

Change Summary:
---------------
* Updated validation.ts
* Made the priorExperience length consistent with the validation

Dev-Ops
=======
- [ ] webpack-config.js

-----------------------

### DIFF
--- a/src/views/Portal/utils/validation.ts
+++ b/src/views/Portal/utils/validation.ts
@@ -49,7 +49,7 @@ export const validateContactForm = (
         } else if (fname.length > 25) {
           isValid = false
           updateErrorMessage(
-            "First Name is Too Long",
+            "Input has 25 char limit",
             "fnameErr",
             dispatchError
           )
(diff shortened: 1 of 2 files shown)


src/views/Portal/utils/validation.ts
```diff
@@ -359,7 +363,7 @@ export const validatedemographicForm = (
         } else if (eventLocation.length > 100) {
           isValid = false
           updateErrorMessage(
-            "Inputted Location Length is too Long",
+            "Response has 100 char limit",
             "eventLocationErr",
             dispatchError
           )
```
Alperdec, line 369: Can we refer to eventLocation in our error message on line 383 as "Event Location"?

Alperdec (approved):
The new error messages look good and are really consistent. Some of our older error messages are inconsistently typed and should probably be revised (capital letters on some nouns/verbs and not others). I can handle this either on this PR or on a separate one.

src/views/Portal/utils/validation.ts
Alperdec: sorry for the code reference, I can't comment on code not added in this PR

Alperdec (approved):
Looks great :)

PR #119, opened 2021-11-16
fixed navbar issues

rvaidun:
Problem
=======
Navbar was having bugs at 1200px
Bug

The sponor us link in the navbar should only show up on the `\` path

Solution
========
* Before mobile view was only rendering if the width was less than 1200px so I changed it to be less than or equal to
* Check the path before rendering the sponsor us link


krishnakpandian (approved):
LGTM

PR #120, opened 2021-11-18
Portal: Track farthest page visited

w-gao:
Problem
=======
Track Farthest Page Visited [Jira](https://cruzhacks-dev.atlassian.net/browse/C2D-172)

Solution
========
* Save progress (max visited page) and add progress state
* Also add "IN PROGRESS" status message in the status component

Change Summary:
---------------
* Display the farthest page visited in the progress bar

Dev-Ops
=======
- [ ] webpack-config.js

-----------------------

### DIFF
--- a/src/views/Portal/components/ApplicationForm/index.view.tsx
+++ b/src/views/Portal/components/ApplicationForm/index.view.tsx
@@ -31,6 +31,7 @@ const ApplicationForm: React.FC = () => {
     setAppStatus,
     accessToken,
     progress,
+    setProgress,
     contactFormData,
     setContactFormData,
     demographicFormData,
@@ -94,6 +95,8 @@ const ApplicationForm: React.FC = () => {
   const { user } = useAuth0()
   const submitData = async () => {
     try {
+      setProgress(ApplicationPages.MLH) // this won't be saved
+
       setSubmitting(true)
       const bodyData = new FormData()
       bodyData.append("fname", contactFormData.fname)
--- a/src/views/Portal/components/ApplicationForm/index.view.tsx
+++ b/src/views/Portal/components/ApplicationForm/index.view.tsx
@@ -6,6 +6,7 @@ import { useApplication } from "components/ApplicationContext/ApplicationContext
 import ApplicationPages from "Props/portal/page"
 import AppStatus from "Props/portal/application"
 import ProgressBar from "components/ProgressBar/ProgressBar"
+import { removeCache } from "utils/Storage"
 import ConnectedForm from "./Connected/index.view"
 import ContactForm from "./Contact/index.view"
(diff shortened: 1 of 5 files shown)


krishnakpandian (approved):
LGTM

krishnakpandian (commented):
I think you should update progress on Submit to the last page if it succeeds so on browser refresh on submitted page it shows that the progress is completed while loading.

w-gao:
Ah, I have to try it out to see what you mean, but I will definitely do that. Great suggestion!

> I think you should update progress on Submit to the last page if it succeeds so on browser refresh on submitted page it shows that the progress is completed while loading.

src/views/Portal/components/ApplicationForm/index.view.tsx
```diff
@@ -93,6 +95,8 @@ const ApplicationForm: React.FC = () => {
   const { user } = useAuth0()
   const submitData = async () => {
     try {
+      setProgress(ApplicationPages.MLH) // this won't be saved
```
krishnakpandian: I think we need to run the save function here to update the page state.

If you want to replicate the bug, submit an application. This will return you to page 0. Refresh the browser, you'll notice while it's loading the bar is not filled for a few seconds and that's because session storage has the incorrect value in the page state.

krishnakpandian (changes requested):
Almost there, just one last change/clarification on the UI bug I mentioned before.

src/views/Portal/components/ApplicationForm/index.view.tsx
w-gao: Ah, yeah, I didn't want to save the progress here since the MLH responses are not saved. I though the progress would be set to 100% if the application is submitted though.

I will take a look and try to fix this tomorrow.

w-gao: @krishnakpandian Updated to update progress on submit.

w-gao:
The problem with 71a6486 now is that if the submission failed, the progress would be at 100%, which might be fine since the user already submitted once, so it's technically complete. But if we want it to stay at ~80%, we can revert 71a6486.

krishnakpandian:
> The problem with [71a6486](https://github.com/CruzHacks/cruzhacks-2022-website/commit/71a6486b8df7cbffc595410e899d071c981422d3) now is that if the submission failed, the progress would be at 100%, which might be fine since the user already submitted once, so it's technically complete. But if we want it to stay at ~80%, we can revert [71a6486](https://github.com/CruzHacks/cruzhacks-2022-website/commit/71a6486b8df7cbffc595410e899d071c981422d3).

Ah okay, that's good to know. I tested it and it seems like it's working as expected so I think it should be good!

krishnakpandian (approved):

PR #121, opened 2021-11-18
Notify on leave if app is started

w-gao:
Problem
=======
On leave of page, notify the user before they leave the page if there are any unsaved changes [Jira](https://cruzhacks-dev.atlassian.net/browse/C2D-168)

Solution
========
* Add a `newChanges` ref to track if there are unsaved changes
  * useState doesn't work well with the unload event, but useRef seems to work

Change Summary:
---------------
* Notify if there are unsaved changes [5dd6af0]

Dev-Ops
=======
- [ ] webpack-config.js

-----------------------


src/views/Portal/components/ApplicationForm/ShortAnswer/index.view.tsx
```diff
@@ -13,12 +14,13 @@ const ShortAnswerPage: React.FC = () => {
...
               name='Why do you want to attend CruzHacks?'
-              handleChange={(e: React.ChangeEvent<HTMLInputElement>) =>
+              handleChange={(e: React.ChangeEvent<HTMLInputElement>) => {
                 setShortAnswerFormData(prev => ({
                   ...prev,
                   whyCruzHacks: e.target.value,
                 }))
-              }
+                setNewChanges()
+              }}
```
krishnakpandian, line 23: I think functionally it works but, it's a little inconsistent from my testing. It seems like this will only run onChange of the textContainer used. I think `setNewChanges` would need to be set on every input component.

krishnakpandian (changes requested):
I think it's functioning how we would want it but the event listener doesn't seem to be run all the time with changes. My guess is, that the saveNewChanges() needs to be applied to every component

src/views/Portal/components/ApplicationForm/ShortAnswer/index.view.tsx
w-gao: Ooops, that’s my bad. I thought this was the `handleChange` function like the other ones. Good catch!

w-gao:
I made some more changes since the last review. Primarily, I made it so it'll display the confirmation popup when new changes are made, even if they are "saved", so user won't accidentally close the tab and lose everything in session storage.

krishnakpandian (approved):
LGTM

PR #122, opened 2021-11-18
Portal: Add loading indicator

w-gao:
Problem
=======
When submitting, we need some indication that the request is taking place [Jira](https://cruzhacks-dev.atlassian.net/browse/C2D-169/)

Solution
========
* Add a "Submitting..." message and change button text to loading

Change Summary:
---------------
* Add some feedback when submitting [f088164]
* 
Dev-Ops
=======
- [ ] webpack-config.js

-----------------------

### DIFF
--- a/src/views/Portal/components/ApplicationForm/index.scss
+++ b/src/views/Portal/components/ApplicationForm/index.scss
@@ -56,6 +56,7 @@
     display: flex;
     justify-content: center;
     @include typography.form-prompts;
+    color: #00b300;
   }
   &__response-err {
     display: flex;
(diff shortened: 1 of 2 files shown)


w-gao:
@krishnakpandian Let me know if this is what you had in mind. I was hoping to make it cooler but it turns out I don't know how to do that. 😅

krishnakpandian (approved):
I think this is good for our needs of just indicating to the user that the app is being processed. It's mainly important in the situation that it's cold started and takes 5-6 seconds to submit. In terms of styling maybe having the ... render every second . -> .. -> ... could be a cool little addition but it's not something I'm worried if we don't have time.

PR #127, opened 2021-11-23
Updating Options, truncating string length

krishnakpandian:
Problem
=======
The option of Transgender should be inclusive of non-binary and needs to be apparent in the option selected.

Solution
========
* Update the Option to Include Non-binary

Change Summary:
---------------
* Updated Checkbox to ensure that long inputs are truncated
* Updated Option Choice in the DemographicPage

Dev-Ops
=======
- [ ] webpack-config.js

-----------------------


Alperdec (approved):
Looks good to me

PR #132, opened 2021-11-25
Need to have navbar removed during initial release

krishnakpandian:
Problem
=======
Initial Release Should not Contain apply button so users aren't inclined to test the application system before configs are tested.

Solution
========
* Removing it temporarily

Change Summary:
---------------
* Updating Navbar Props

Dev-Ops
=======
- [ ] webpack-config.js

-----------------------


w-gao (approved):
I don't think it's necessary but should we also comment out the /portal route?

krishnakpandian:
> I don't think it's necessary but should we also comment out the /portal route?

I want to do 1 test submission with the production configs without having users navigating to that route.

PR #134, opened 2021-11-29
Updating MLH page and universities

krishnakpandian:
Problem
=======
MLH page is confusing with optional choices that have no impact. MLH also wants naming to be from a pre-chosen list of schools.

Solution
========
* Updated Universities using new list
* Updated MLH page to remove secondary Options

Change Summary:
---------------
* Updated MLHPage.tsx
* Updated Utils/universities.js

Dev-Ops
=======
- [ ] webpack-config.js

-----------------------
* Presidents decided let's keep the invalid options despite client side errors

### DIFF
--- a/src/views/Portal/components/ApplicationForm/MLH/index.view.tsx
+++ b/src/views/Portal/components/ApplicationForm/MLH/index.view.tsx
@@ -69,9 +69,6 @@ const MLHPage: React.FC = () => {
             {
               label: "I have read and agree to the terms outlined above.",
             },
-            {
-              label: "I do not agree to the terms above.",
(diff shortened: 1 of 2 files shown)


Alperdec (approved):
This looks good, although it's going to cause validation errors. We can open up two separate bug branches for validation on both client and backend

krishnakpandian:
@Alperdec I talked with the presidents about this, we aren't going to update the validation because this can lead to a large amount inputs we have not considered. We were torn between removing those schools or keeping them in there just for MLH sake even if it may error client side but we decided to leave it there even if it errors client side.

PR #139, opened 2021-12-01
Only show unsaved changes prompt if app is in progress

w-gao:
Problem
=======

After Submitting, when closing the tab the unsaved changes prompt still runs [Jira](https://cruzhacks-dev.atlassian.net/browse/C2D-178)

Solution
========
* ~~Add status dependency to the useEffect hook for the unsaved changes prompt~~
* Move unsaved changes check to `ApplicationForm`

Change Summary:
---------------

Dev-Ops
=======
- [ ] webpack-config.js

-----------------------

### DIFF
--- a/src/components/ApplicationContext/ApplicationContext.tsx
+++ b/src/components/ApplicationContext/ApplicationContext.tsx
@@ -42,6 +42,7 @@ interface ApplicationContextProps {
   savePage: any
 
   // track if user has inputted data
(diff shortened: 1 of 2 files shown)


w-gao (commented):
Now that I think about it maybe this should belong in the `ApplicationForm` component?

krishnakpandian (approved):
Tested and it seems like this works!

PR #144, opened 2021-12-10
fixed progress bar bug

sidhantas:
Problem
=======
https://cruzhacks-dev.atlassian.net/browse/C2D-177?atlOrigin=eyJpIjoiZDQ3ZmExNmNjNTVjNGJiN2E3MWYzOGM0MDJhNDY4MWIiLCJwIjoiaiJ9

Solution
========
*  Changed progress bar to be filled on accepted/rejected status

Change Summary:
---------------
* Something else I/we did

Dev-Ops
=======
- [ ] webpack-config.js

-----------------------

### DIFF
--- a/src/components/ApplicationContext/ApplicationContext.tsx
+++ b/src/components/ApplicationContext/ApplicationContext.tsx
@@ -147,7 +147,6 @@ export const ApplicationProvider: React.FC = () => {
                 12 * 60 * 60, // store for 12hrs
                 user?.email
               )
-              setProgress(6)
               break
             case "rejected":
               setStatus(AppStatus.Rejected)
@@ -157,7 +156,6 @@ export const ApplicationProvider: React.FC = () => {
                 12 * 60 * 60, // store for 12hrs
                 user?.email
               )
-              setProgress(6)
               break
             case "pending":
               setStatus(AppStatus.Pending)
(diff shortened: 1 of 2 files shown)


src/components/ApplicationContext/ApplicationContext.tsx
```diff
@@ -147,6 +147,7 @@ export const ApplicationProvider: React.FC = () => {
                 12 * 60 * 60, // store for 12hrs
                 user?.email
               )
+              setProgress(6)
```
krishnakpandian: Let's use Macro for ApplicationPages when possible.

src/components/ApplicationContext/ApplicationContext.tsx
```diff
@@ -156,6 +157,7 @@ export const ApplicationProvider: React.FC = () => {
...
```
krishnakpandian: Same as above

krishnakpandian (changes requested):
For the submitted state it's implemented a little differently but your way is completely fine. Just for consistency, can we have the submitted state also implemented the same way. You can choose to keep it your way or use the previous implementation. Here's how it's currently implemented: https://github.com/CruzHacks/cruzhacks-2022-website/blob/30aba136d2ce0acd704723cda36adfbba067176a/src/views/Portal/components/ApplicationForm/MainPage/index.view.tsx#L48-L50

Also once you do that, let's fill out the PR Form just so we can have reference if we need to review anything in the future.

krishnakpandian (approved):
LGTM

PR #146, opened 2021-12-14
Updating Error Message to Tell user to Refresh Browser

krishnakpandian:
Problem
=======
Some users are unable to submit because auth tokens are expiring due to their session timing out and may not know to refresh their browser.

Solution
========
* Updated the error message Text to tell user to refresh browser
* Rerequesting token if session is still available

Change Summary:
---------------
* Updated ApplicationForm

Dev-Ops
=======
- [ ] webpack-config.js

-----------------------

### DIFF
--- a/src/views/Portal/components/ApplicationForm/index.view.tsx
+++ b/src/views/Portal/components/ApplicationForm/index.view.tsx
@@ -33,6 +33,7 @@ const ApplicationForm: React.FC = () => {
     setSubmitting,
     setAppStatus,
     accessToken,
+    setToken,
     progress,
     contactFormData,
     setContactFormData,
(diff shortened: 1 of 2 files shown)


src/views/Portal/components/ApplicationForm/index.view.tsx
```diff
@@ -199,6 +200,11 @@ const ApplicationForm: React.FC = () => {
       } else {
         setServerErrors([])
       }
+      if (err && err.response && err.response.status === 401) {
```
w-gao, line 203: Uh oh.. How often does the token expire?

w-gao (approved):

src/views/Portal/components/ApplicationForm/index.view.tsx
krishnakpandian: It expires after 24 hours which is fairly forgiving. I've only seen ~3 instances of users getting 401s according to our loggers so it's not a huge issue but definitely something worth clarifying over the last day of applications.

PR #152, opened 2021-12-23
Speakers component

w-gao:
Problem
=======

Speakers component [Jira](https://cruzhacks-dev.atlassian.net/browse/C2D-73)

Solution
========
* Implement Speakers compoennt

Change Summary:
---------------

Dev-Ops
=======
- [ ] webpack-config.js

-----------------------

This should mostly be responsive, but may need tweaking for certain screen sizes.


src/Props/speakers/cards.ts
```diff
@@ -0,0 +1,55 @@
+import { CardProps } from "Props/props"
+import images from "./images"
+
+const speakerCards: CardProps[] = [
+  {
+    name: "Antonio Xu Liu",
+    title: "Senior Strategy Program Manager\n@ National Grid",
+    imgSrc: images.seaOtter,
+    LinkedInLink: "",
```
w-gao, line 9: If we need to include more information about the speakers, I can make another interface for this. Right now I'm using the same interface as our team cards, since they are almost the same.

src/views/Home/components/Speakers/index.scss
```diff
@@ -0,0 +1,67 @@
...
+      background: #ffffff;
+      width: fit-content;
+
+      @include typography.header-text;
+      font-size: 52px;
+      color: #334276;
+
+      @include responsiveness.screen(smallest, phone, tablet) {
+        font-size: 40px;
+      }
```
krishnakpandian: For screen sizes under 360px it might be better to try using a different font-size. In the example below it shows a bit bulkier header. I find that size 30px seems to work nicely for the case of 280px.

speakers-smallest
.

src/views/Home/components/Speakers/SpeakerCard/SpeakerCard.tsx
```diff
@@ -0,0 +1,30 @@
...
+      <ImgOverlay className='speaker-card__img-overlay' />
+      <img src={imgSrc} alt={name} className='speaker-card__image' />
+    </div>
+    <div className='speaker-card__card-container'>
+      <div className='speaker-card__name'>{name}</div>
+      <div className='speaker-card__title'>
+        {title.split("\n").map(val => (
+          <>
+            {val}
+            <br />
```
krishnakpandian: I'm not the biggest fan of using br tags. Overall it looks good but do you think there's another way we can implement this consistently. I know realistically `<div>` should be able to work but if this works best we can keep it like this.

krishnakpandian (changes requested):
Overall looks really good just a couple of quick changes! We can merge this in once we get the remaining information we need for each speaker

src/views/Home/components/Speakers/SpeakerCard/SpeakerCard.tsx
w-gao: Ah yeah, I decided to add this last minute so this can probably be implemented better. It looks like `<div>`s work pretty well so I'll use that instead.

krishnakpandian (commented):
Looks good! I'll approve once we get the images and links for the speakers. I'll update you once we get that as well

Alperdec (commented):
it looks good to me, however is there any way we can get the border rings around the speaker's photo to be less pixelated?

w-gao:
> it looks good to me, however is there any way we can get the border rings around the speaker's photo to be less pixelated?

Ah, good catch! I didn't even notice. Looks like it's the `shape-rendering="crispEdges"` property in the svg that caused it. Hopefully this should look better now.

krishnakpandian:
@w-gao So I just got the linkedin information for all the speakers. We'll be using their linkedIn Photos for our site. This [document](https://docs.google.com/document/d/1GUwyWeQC2oDsTViqqY1QIh0nVFyYWihZ53pFSDcGyMA/edit) should contain all the linkedin information to get the images. Let me know if you need anything else!

w-gao:
> @w-gao So I just got the linkedin information for all the speakers. We'll be using their linkedIn Photos for our site. This [document](https://docs.google.com/document/d/1GUwyWeQC2oDsTViqqY1QIh0nVFyYWihZ53pFSDcGyMA/edit) should contain all the linkedin information to get the images. Let me know if you need anything else!

Great! I added their images from linkedIn.

krishnakpandian (approved):
Looks Great! I think the only last change I would make is maybe removing/shortening the underline on widescreens. Other than that, it looks good to me!

speakers-widescreen

w-gao:
@krishnakpandian Yeah, good point. I removed it on widescreens - let me know how this looks now.

krishnakpandian:
@w-gao it looks good! I think we can ship it!

PR #154, opened 2021-12-24
Simplfying Parent Container for Puzzle

krishnakpandian:
Problem
Let's try Keeping the Puzzle as Modular as possible so after the event we can remove it quickly if needed and duplicate it quickly as well.

Change Summary:
---------------
* Moved State Variable to scope of Puzzle Function
* Updated Props to pass in component as children

Dev-Ops
=======
- [ ] webpack-config.js

-----------------------


Alperdec (approved):
Thanks for the changes!

PR #155, opened 2021-12-24
Feature/tracks

sidhantas:
Problem
=======
https://cruzhacks-dev.atlassian.net/browse/C2D-29?atlOrigin=eyJpIjoiYjg0ZGRmY2QwZjJhNDZlMGE2NTI3ODVjNzdlZjk0ZWIiLCJwIjoiaiJ9

Solution
========
* Created the tracks component

Change Summary:
---------------
* Added Prize Tracks to home screen
* Created hover effect for prize tracks

Dev-Ops
=======
- [ ] webpack-config.js

-----------------------


src/views/Home/components/PrizeTracks/PrizeTracks.tsx
```diff
@@ -0,0 +1,63 @@
+import * as React from "react"
+import "./PrizeTracks.scss"
+import HealthHacks from "./SVGS/HealthHacks"
+import BeginnerHacks from "./SVGS/BeginnerHacks"
+import EarthHacks from "./SVGS/EarthHacks"
+import GoldenStateHacks from "./SVGS/GoldenStateHacks"
+import JusticeHacks from "./SVGS/JusticeHacks"
+import UIUXHacks from "./SVGS/UIUXHacks"
```
krishnakpandian: For SVGs there's a different way you can implement this which I think is much cleaner and simpler than creating a component for each of these. With an svg file you can write `import { ReactComponent as HealthHacks } from "path/to/svg"` and do a similar thing to what you did here. You wouldn't need to create all these tsx files and it's a lot quicker. I'll link an example of this method from our codebase if you need [reference](https://github.com/CruzHacks/cruzhacks-2022-website/blob/e16cd59de9b19ca089b5ef0113971c72a898c17b/src/components/Footer/Footer.tsx#L5-L7).

krishnakpandian:
image

I think the emulator SASS isn't accurate to how it's being run on mobile devices. You may need to update how the circle is functioning.

src/views/Home/components/PrizeTracks/PrizeTracks.scss
```diff
@@ -0,0 +1,67 @@
...
+        width: inherit;
+        aspect-ratio: 1/1;
+        position: absolute;
+        z-index: 100;
+        display: flex;
+        justify-content: center;
+        align-items: center;
+        @include typography.title;
+        font-style: italic;
+        text-align: center;
```
krishnakpandian: tracks-smaller-phones

On smaller screens, I've noticed when you click the image to show the circle outline and the size is exceeding length of the screen.

src/views/Home/components/PrizeTracks/PrizeTracks.tsx
```diff
@@ -0,0 +1,63 @@
...
+    </div>
+    <div className='prizeTracks__images'>
+      {Tracks.map(image => (
+        <Circle
+          image={image[0]}
+          hoverText={image[1] as string}
+          key={image[2] as number}
+        />
+      ))}
+    </div>
```
krishnakpandian, line 83: Putting a max-width can help have the spacing scale for screen sizes > 1600px.
Ig Widescreen:

tracks-large-screen

I would suggest setting a max-width that will maintain 3 per row after 1600px but whatever scales best.

krishnakpandian (changes requested):
Just a couple of changes, mainly with the svgs and the hover effect. Overall it looks good!

src/views/Home/components/PrizeTracks/PrizeTracks.scss
sidhantas: Can I ask how you're displaying the content on your phone? I tried using ngrok but it seems like auth0 doesn't like that

krishnakpandian: I've seen some people have issues with auth0 on mobile devices still. This is my setup but I'm not sure if this will work for everyone. 

I use a windows machine + iPhone, on my windows machine I'm connected to home Wi-Fi and I have it set as a private network in my laptop settings. On my phone, I am connected to the same Wi-Fi and I use the network address given from React console after the server has been setup. On my phone, non-authenticated routes seems to work well for me but I know some people have had issues with auth0 still. If you want, you can comment out the Auth0 Provider within `src/auth/Auth0ProviderWithHistory.tsx` and that should work for non-authenticated routes.

In the end, we still run it on the development branch to catch these issues since I know it's difficult to test for everyone.

sidhantas: Sounds good, I think I fixed it but I only have an android so I cant test for apple rendering

src/views/Home/components/PrizeTracks/PrizeTracks.tsx
```diff
@@ -0,0 +1,87 @@
...
+import "./PrizeTracks.scss"
+import { ReactComponent as HealthHacks } from "images/tracks/health_hacks.svg"
+import { ReactComponent as BeginnerHacks } from "images/tracks/beginner_hacks.svg"
+import { ReactComponent as EarthHacks } from "images/tracks/earth_hacks.svg"
+import { ReactComponent as GoldenStateHacks } from "images/tracks/golden_state_hacks.svg"
+import { ReactComponent as JusticeHacks } from "images/tracks/justice_hacks.svg"
+import { ReactComponent as UIUXHacks } from "images/tracks/ui_ux_hacks.svg"
+
+interface CircleProps {
+  image: any
```
krishnakpandian, line 11: Image can be of type React.ReactNode if I'm correct.

src/views/Home/components/PrizeTracks/PrizeTracks.scss
```diff
@@ -0,0 +1,107 @@
...
+  -o-user-select: none;
+  user-select: none;
+}
+
+.UIUX {
+  margin-left: 35px;
+  margin-top: 25px;
+}
+
+.svg-mask {
```
krishnakpandian: Can we place these classes inside prize tracks just so we can reduce naming collisions

src/views/Home/components/PrizeTracks/PrizeTracks.scss
```diff
@@ -0,0 +1,107 @@
+@use "~stylesheets/typography";
+
+.prizeTracks {
+  background: linear-gradient(#b0b6f7, #b0b6f7);
```
krishnakpandian: I think the background is currently set to what the component above and below in this build. That's going to change once we get the other components merged in so I would update this based on the figma colors.

krishnakpandian: Let's add a margin top of -1px to get rid of the invisible line.

krishnakpandian (changes requested):
Looks great, just some finishing touches!

src/views/Home/components/PrizeTracks/PrizeTracks.scss
```diff
@@ -0,0 +1,107 @@
...
+    flex-wrap: wrap;
+    justify-content: center;
+    width: 100%;
+    max-width: 1600px;
+    align-items: center;
+    &__container {
+      margin: 40px 50px;
+      width: 341px;
+      min-width: 260px;
+      aspect-ratio: 1/1;
```
krishnakpandian: It's still not displaying on mobile correctly but I think it's because of aspect ratio not having a lot of browser support.

krishnakpandian:
```suggestion
      min-height: 260px;
```

src/views/Home/components/PrizeTracks/PrizeTracks.scss
```diff
@@ -0,0 +1,107 @@
...
+      align-items: center;
+      position: relative;
+      box-shadow: -4px 8px 4px rgba(20, 28, 94, 0.4); 
+      &--hover {
+        background-color: #162C7B;
+        opacity: 0.76;
+        color: white;
+        border-radius: 50%;
+        width: 100%;
+        aspect-ratio: 1/1;
```
krishnakpandian:
```suggestion
        height: 100%;
```

src/views/Home/components/PrizeTracks/PrizeTracks.tsx
```diff
@@ -0,0 +1,87 @@
...
+    3,
+  ],
+  [
+    <GoldenStateHacks className='prizeTracks--track-img-dimensions-golden' />,
+    "Golden State Hacks",
+    4,
+  ],
+  [
+    <UIUXHacks className='prizeTracks--svg-shadow prizeTracks--track-img-dimensions UIUX' />,
+    "UI UX Hacks",
```
krishnakpandian: It should UI / UX Hacks

src/views/Home/components/PrizeTracks/PrizeTracks.tsx
```diff
@@ -0,0 +1,87 @@
...
+import { ReactComponent as GoldenStateHacks } from "images/tracks/golden_state_hacks.svg"
+import { ReactComponent as JusticeHacks } from "images/tracks/justice_hacks.svg"
+import { ReactComponent as UIUXHacks } from "images/tracks/ui_ux_hacks.svg"
+
+interface CircleProps {
+  image: any
+  hoverText: string
+}
+
+const Tracks = [
```
krishnakpandian, line 15: For keys in React, generally we don't want to use indexes. Since this is a hard-coded array it's fine but just for reference in the future. Also here's a quick [article](https://robinpokorny.medium.com/index-as-a-key-is-an-anti-pattern-e0349aece318) if you want to check it out explaining it much more in-depth.

krishnakpandian (approved):
LGTM

src/views/Home/components/PrizeTracks/PrizeTracks.tsx
w-gao: 👍

w-gao (approved):
Other than @krishnakpandian's comments, this looks great to me!

krishnakpandian:
@sidhantas Can we resolve the merge conflicts and then "UI / UX Hacks" change so we can merge into development.

PR #158, opened 2021-12-28
Schedule component

w-gao:
Problem
=======
Schedule component [Jira](https://cruzhacks-dev.atlassian.net/browse/C2D-74)

Solution
========
* Create Schedule component

Change Summary:
---------------
* Schedule component layout on desktop d18b166
* Responsiveness 3e8559f

Dev-Ops
=======
- [ ] webpack-config.js

-----------------------

* This should probably be merged after #152 and #155; I need to fix the background gradient

* Desktop

* Tablet

* Mobile

  - NOTE: should ask design about the arrows on mobile


Alperdec (changes requested):
Looks really good William! The only comments I have is comparing to figma there seems to be a little more padding between the text and the table fields and the background gradient differences. It looks like we'll have to update the gradient for the FAQ component as well.

src/views/Home/components/Schedule/components/index.scss
```diff
@@ -0,0 +1,118 @@
...
+    color: #d3ddff;
+    font-size: 20px;
+    width: 100%;
+
+    @include responsiveness.screen(smallest, phone, tablet) {
+      font-size: 16px;
+    }
+
+    &__time {
+      margin: 0;
```
krishnakpandian: On desktop, I think the margin should be a little bit higher according to the figma

src/views/Home/components/Schedule/components/index.scss
```diff
@@ -0,0 +1,118 @@
...
+
+      @include responsiveness.screen(smallest) {
+        padding: 8px 5px;
+        width: 75px;
+        min-width: 75px;
+      }
+    }
+
+    &__text {
+      margin: 0;
```
krishnakpandian: Same as time, for desktop, the margin should be a bit higher

src/views/Home/components/Schedule/components/index.scss
```diff
@@ -0,0 +1,118 @@
+@use "./stylesheets/responsiveness";
+@use "./stylesheets/typography";
+
+.schedule-table {
+  position: relative;
+  width: 100%;
+
+  &__table {
+    min-height: 300px;
```
krishnakpandian: From the figma, it seems like the table should have the same height for each component and the min-height should be a bit higher on desktop

krishnakpandian (changes requested):
This looks great! I agree with @Alperdec comments. I think the main thing is that on desktop the vertical sizing is a bit off but overall it looks good.

w-gao:
Thanks for the review @Alperdec and @krishnakpandian! I adjusted the vertical spacing and made each component the same height, so it should look more like the figma design now.

src/views/Home/components/Schedule/components/index.scss
```diff
@@ -0,0 +1,140 @@
...
+    min-height: 500px;
+    height: 500px;
+  }
+
+  &__event {
+    display: flex;
+    flex-wrap: nowrap;
+    border-bottom: 2px solid #9fc5ff;
+    color: #d3ddff;
+    font-size: 20px;
```
krishnakpandian, line 61: Font-size should be closer to 30px on desktop + widescreens. We can use our font macro or screen macro for a quick fix.

krishnakpandian, line 61: Font-size should be closer to 30px on desktop + widescreens. We can use our font macro or screen macro for a quick fix.

src/views/Home/components/Schedule/components/index.scss
```diff
@@ -0,0 +1,140 @@
...
+    width: 100%;
+    height: 100%;
+    max-height: 60px;
```
krishnakpandian, line 64: This seems like it should be closer to 80px on desktop + widescreen.

src/views/Home/components/Schedule/components/index.scss
```diff
@@ -0,0 +1,140 @@
...
+      font-size: 20px;
+    }
+  }
+
+  &__container {
+    display: flex;
+    flex-direction: column;
+    align-items: stretch;
+    min-height: 500px;
+    height: 500px;
```
krishnakpandian, line 49: These seem closer to 800px for desktop.

krishnakpandian (changes requested):
I think we are almost there, it still seems a bit rectangular in comparison to the figma for desktop views. The mobile views looks amazing!

krishnakpandian (approved):
LGTM

